PR #7, opened 2018-03-15
Ci build

ehildenb:
Do not merge yet, testing that CI setup works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,22 @@
 # --------
 
 specs_dir:=specs
+build_dir:=.build
+
+.PHONY: all clean
 
 all: k-files split-proof-tests
 
 clean:
-	rm -rf $(specs_dir) .build/
+	rm -rf $(specs_dir) $(build_dir)
+
+pandoc_tangle_submodule:=$(build_dir)/pandoc-tangle
+TANGLER:=$(pandoc_tangle_submodule)/tangle.lua
+LUA_PATH:=$(pandoc_tangle_submodule)/?.lua;;
+export LUA_PATH
+
+$(TANGLER):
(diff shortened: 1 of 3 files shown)


ehildenb:
@daejunpark ready for review now.

Makefile
```diff
@@ -52,65 +99,68 @@ hobby-erc20: $(patsubst %, $(specs_dir)/hobby-erc20/%, $(hobby_erc20_files)) $(s
...
-	pandoc --from markdown --to "$(TANGLER)" --metadata=code:".k" $< > $@
-
-# Spec Files
-# ----------
-
 # Bihu
-$(specs_dir)/bihu/collectToken-spec.k: bihu/module-tmpl.k bihu/spec-tmpl.k bihu/collectToken-spec.ini
+bihu_tmpls:=bihu/module-tmpl.k bihu/spec-tmpl.k
+
+$(specs_dir)/bihu/collectToken-spec.k: bihu/collectToken-spec.ini $(TANGLER)
```
daejunpark: why did you drop the tmpls from the dependency list?

ehildenb: Well, if they're static, then I think they don't belong in the prereqs. If they're generated, then they do. My impression was that they're static.

The reason I did is is because I wanted to make these targets depend on the TANGLER, but now I realize that they actually don't depend on the TANGLER. I will remove that dependency and force push.

ehildenb:
@daejunpark I've undone the mistake you pointed out.

daejunpark:
looks good. i noticed that `kevm prove` is used instead of `kevm test`. but it seems to be ok, as the negative test #8 failed.

PR #10, opened 2018-03-23
erc20/, resources: switch #exception => #exception EVM_INVALID_INSTRU…

ehildenb:
…CTION

@daejunpark I have switched over to using the typed exceptions. This won't even parse correctly on CI until https://github.com/kframework/evm-semantics/pull/167 is merged into KEVM.

The thing I'm not understanding though: the tests pass on https://github.com/kframework/evm-semantics/pull/167, but I thought the Vyper compiler does not generate `INVALID` opcode, and I switched all the proofs over to using the `INVALID_INSTRUCTION` exception type. Any insight into why the proofs are passing?


ehildenb:
Obsolete

PR #11, opened 2018-03-28
Fix typo in gen: sepc_template -> spec_template

ryanberryhill:
In gen(), the parameter sepc_template is a misspelling of spec_template. It happens to work because the value of the global spec_template (which is passed to gen() as sepc_template) on line 77 is used instead inside of gen().


daejunpark:
@ryanberryhill good catch! thanks for spotting and fixing this!

PR #16, opened 2018-04-24
casper/casper-spec.ini: format the spec

yzhang90:


yzhang90:
@daejunpark yes,please

PR #19, opened 2018-05-02
Vote proc reward

yzhang90:
link proc_reward with vote


daejunpark:
merge if https://github.com/kframework/evm-semantics/pull/209 passes

PR #39, opened 2018-07-27
[WIP] packed storage

tomtomjhj:


Makefile
```diff
@@ -185,7 +271,8 @@ gnosis_files:=setup-spec.k \
 #             getTransactionHash-spec.k
 #             checkHash-spec.k
 
-proof_tests:=bihu vyper-erc20 zeppelin-erc20 hkg-erc20 hobby-erc20 sum-to-n ds-token-erc20 gnosis
+#proof_tests:=bihu vyper-erc20 zeppelin-erc20 hkg-erc20 hobby-erc20 sum-to-n ds-token-erc20 gnosis zilliqa_erc20 #tron_erc20 icon_erc20 zilliqa_erc20 zrx_erc20
+proof_tests:=zilliqa_erc20 #tron_erc20 icon_erc20 zilliqa_erc20 zrx_erc20
```
daejunpark: why did you this? we should test all the existing ones (except the casper).

daejunpark:
@tomtomjhj please split this big PR into smaller pieces.

my suggestion.
1. the packed storage abstraction and lemma only
2. first token contract code and spec
3. 2nd token contract code and spec
4. ...
N+1. Nth token contract code and spec

Makefile
tomtomjhj: for a quick test

erc20/verification.k
```diff
@@ -4,4 +4,128 @@ requires "../lemmas.k"
...
+
+ // |Int
+ // 1. simbolic I, zero offset
+    rule I |Int #stWord(0 VS, L LS) => #stWord(I VS, L LS)
+      requires 0 <=Int I andBool I <=Int (256 ^Int L) -Int 1 andBool notBool #isConcrete(I)
+
+ // 2. symbolic I, non zero offset
+    rule (I *Int OFFSET_POW) |Int #stWord(VS, LS)
+      => #stWord(#updateIntList(VS, I, log256Int(OFFSET_POW), LS), LS)
+      requires OFFSET_POW ==Int 256 ^Int log256Int(OFFSET_POW)
```
daejunpark: @tomtomjhj why did you separate two cases here? i mean, i think it is better to merge two cases into a general one, unless it doesn't incur performance overhead too much. the readability is more important than the performance here.

tomtomjhj: Essentially cases 1 and 2 can be merged if I make a helper function that returns offset of `I` or `I *Int OFFSET_POW`.
Case 3 is different from 1 and 2 because the offset cannot be retrieved from a concrete value.

PR #43, opened 2018-08-08
erc20/verification: #rangeBool, lemmas for bool variables, and overflow 

tomtomjhj:


erc20/verification.k
```diff
@@ -4,4 +4,19 @@ requires "../lemmas.k"
 module VERIFICATION
     imports EDSL
     imports LEMMAS
+
+ // some treatments for solidity bool variables
+    syntax Bool ::= #rangeBool(Int)  [function]
+                  | #isBoolWord(Int) [function, smtlib(isBoolWord)]
```
daejunpark: @tomtomjhj where is the definition of `#isBoolWord`?

tomtomjhj: No rules for `#isBoolWord`. The only purpose of this term is to provide the fact that a symbolic `Int` term is a solidity bool variable, so that the lemmas below can be applied safely.

erc20/verification.k
```diff
@@ -4,4 +4,19 @@ requires "../lemmas.k"
...
+
+    rule #rangeBool(B) => #range(0 <= B <= 1) andBool #isBoolWord(B)
+ // b != false -> b
+    rule bool2Word(notBool(B ==K 0)) => B        requires #isBoolWord(B)
+ // b != true  -> !b
+    rule bool2Word(notBool(B ==K 1)) => 1 -Int B requires #isBoolWord(B)
```
daejunpark: @tomtomjhj what's the problem having `#range(0 <= B <= 1)` instead of `#isBoolWord(B)` in the `requires`?

daejunpark: Or, simply, `#rangeBool(B)`

daejunpark: btw, could you provide an example context that requires this?

tomtomjhj: hmm I don't remember the exact reason I ended up defining #isBoolWord` instead of `#range(0 <= B <= 1)`, but just using `#range` looks perfectly fine. I'll try that tomorrow.

PR #49, opened 2018-09-23
add explicit import

dwightguth:
I believe the absence of this import in this module is what caused the prover to slow down when I made my changes to EVM. K-REFLECTION was no longer getting pulled in, which meant that the lemmas with #isConcrete were no longer being applied, slowing down the proofs.

### DIFF
--- a/resources/lemmas.md
+++ b/resources/lemmas.md
@@ -6,6 +6,7 @@ requires "evm.k"
 
 module LEMMAS
     imports EVM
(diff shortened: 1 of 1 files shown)


dwightguth:
I can confirm that this fixes the bug in performance for kframework/evm-semantics#257. Can we please get this approved and merged? Note that I just enabled required PR approvals because it really should be on for all our repositories.

PR #57, opened 2018-10-18
Formal verification of GNO-ERC20

dteiml:


erc20/gno/README.md
```diff
@@ -0,0 +1,59 @@
+# Formal Verification of GNO ERC20 Token
+
+By Dominik Teiml (dominik@gnosis.pm), [Gnosis](https://www.gnosis.pm)
+
+We formally verified the GNO ERC20 token contract [runtime bytecode](./gno-erc20.bytes).
+
+We found the following deviations from [ERC20-EVM](../vyper/vyper-erc20-spec.ini):
+
+- In the [high-level Solidity code](https://etherscan.io/address/0x6810e776880c02933d47db1b9fc05908e5386b96#code), mapping of users' allowances is called `allowed`, hence we use that in our [eDSL spec](./gno-erc20-spec.ini).
+- In `transfer-failure-1` and `transferFrom-failure-1`, `BAL_TO +Int VALUE >=Int (2 ^Int 256)` was removed from the constraint. The reason is the GNO was constructed with a fixed supply of 10 M. Hence no overflow protections are necessary (and so none are present)
```
daejunpark: I'm not sure about this reasoning, unless you additionally prove that the sum of all balances is equal to the total supply.

erc20/gno/README.md
```diff
@@ -0,0 +1,59 @@
...
+- A new lemma was added to be added to [lemmas.md](../../resources/lemmas.md) to help K tool with the reduction of a specific term. We present the lemma as well as its soundness proof:
+
+## Lemma
+
+Lemma:
+```
+rule chop ( W0:Int +Int W1:Int ) -Word W1:Int => chop ( W0 )
```
daejunpark: Now I see why you need this additional lemma, while it wasn't needed for other ERC20 tokens.

I found that the GnosisToken's transferFrom has:
```
        balances[_to] += _value;
        balances[_from] -= _value;
```
instead of:
```
        balances[_from] -= _value;
        balances[_to] += _value;
```

Even if all your reasoning is correct, I don't see why it is written that way at the first place, while the latter code seems much easier to understand and reason. I mean, one has to think twice when he is trying to understand and make sure it is correct.

daejunpark:
This needs to be rebased to the master branch instead of the gnosis branch.

erc20/gno/README.md
dteiml: I see.  I'll think about the best way for that tomorrow

erc20/gno/README.md
dteiml: I guess the answer is that the author didn't have exposure to formal verification and saw them as equivalent ;) Anyways, I didn't write it so this is what we have. And K fails without the lemma

denis-bogdanas:
@dteiml I just verified your lemma, sorry for delay, too many other things to do.
Lemma needs range conditions for W0 and W1, like `#rangeUInt256(W0)`. It might be correct otherwise too, but I'm not sure of some subtleties of `modInt`, how it operates with negative arguments.

@daejunpark Why there's need for 2 rules for `-Word` in the first place? We had tons of complications because of this.
```
rule W0 -Word W1 => chop( W0 -Int W1 ) requires W0 >=Int W1  
rule W0 -Word W1 => chop( (W0 +Int pow256) -Int W1 ) requires W0 <Int W1
`rule chop ( I:Int ) => I modInt pow256`
```

If `modInt` always returns positive result, then 1st rule should be enough. It should return same result as 2nd rule. What's the exact semantics of `modInt` and how is it different from `%Int`?

PR #58, opened 2018-10-19
specs for testSignatureSplit

yzhang90:


Makefile
```diff
@@ -192,7 +192,10 @@ gnosis_test_files:=testKeccak-data1-spec.k \
                    testKeccak-data33-spec.k \
                    testAbiEncode-spec.k \
                    testAbiEncode-AndKeccak-data1-spec.k \
-                   testAbiEncodePacked-spec.k
+                   testAbiEncodePacked-spec.k \
+		   testSignatureSplit-pos0-spec.k \
+		   testSignatureSplit-pos1-spec.k \
```
daejunpark: nitpick: alignment of the above two lines.

gnosis/test/ApiTest.hex
```diff
@@ -1,4 +1,4 @@
 
-======= gnosis/test/ApiTest.sol:ApiTest =======
+======= ApiTest.sol:ApiTest =======
```
daejunpark: specific reason to change this file path?

@denis-bogdanas what do you think?

daejunpark:
looks good to me except the comments.

@denis-bogdanas could you review this as well?

daejunpark:
@denis-bogdanas I rebased it onto the current master. Please merge if you are happy with it.

gnosis/test/ApiTest.hex
denis-bogdanas: This is a generated file. Path is irrelevant.

gnosis/test/api-test.ini
```diff
@@ -162,9 +162,47 @@ proxy_storage:
...
+    andBool #rangeUInt(256, SIG1_R)
+    andBool #rangeUInt(256, SIG1_S)
+    andBool #rangeUInt(8  , SIG1_V)
+    andBool #rangeUInt(256, SIG2_R)
+    andBool #rangeUInt(256, SIG2_S)
+    andBool #rangeUInt(8  , SIG2_V)
+    andBool #rangeUInt(256, POS)
+
+[testSignatureSplit-pos0]
+output: _ => #encodeArgs(#uint8(SIG0_V), #bytes32(SIG0_R), #bytes32(SIG0_S))
```
denis-bogdanas: The construct `#encodeSig` leads to complications here. As you see, output has to be specified separately for each pos value. A more general approach would be to specify input as #toBytes(DATA, 65*N), and output as a certain bytes range depending on `pos`. Then you could move specification for output to the main spec.

You might need an additional construct for a byte range, like `#byteRange(DATA, START, LEN, N)`.

In addition I'm wondering if it's possible to prove the spec for all pos values <=N, but we can leave this for later.

resources/lemmas.md
```diff
@@ -214,6 +214,44 @@ These rules are specific to reasoning about EVM programs.
     rule N &Int MASK => N  requires MASK ==Int (2 ^Int (log2Int(MASK) +Int 1)) -Int 1 // MASK = 0xffff...f
                             andBool 0 <=Int N andBool N <=Int MASK
 
+
```
denis-bogdanas: No need for side consitions. We assume that elements in a WordStack are already in the required range. Also, if this rule has to stay, it should be made general for any WordStack Length. Look at the rules at the end of `verification.k` for example:
https://github.com/runtimeverification/verified-smart-contracts/blob/8b798a26f7adb30080b6b2aa13e050f36dc403ed/gnosis/verification.k#L75-L84

gnosis/test/api-test.ini
yzhang90: #toBytes(DATA, 65*N) may cause some additional problems for writing specs and symbolic reasoning. First, current java backend doesn't support  looking up symbolic keys from map. we need to enumerate all the possible pos anyway. Second, If we use #toBytes(DATA, 65*N), when loading value from specific memory location, the value will be like #asWord(....) instead of a single value like `SIG0_R`and the output will be very large. If I want to further constraint the SIG0_V, SIG0_R and SIG0_S, instead of directly writing constraint on them, I need to write constraint on three byte arrays.

denis-bogdanas: That's why I proposed to have a byte range construct. Everywhere you have SIG0_V, SIG0_R and SIG0_S now, you would have a certain byte range over DATA instead. And you can constrain it the same way as we constrain data. The advantage is that you can write a single spec, with output fully specified, for ANY N and POS, and restrict LEN and POS in sub-specs byt additional +requires clauses. I guess it would give us more flexibility on the long term in other areas too.

The spec will remain compact. And later we can think of some abstraction that avoids unfolding the byte range into nthbyteof sequences, like the one Daejun did for hash1 and hash2. Then configuration will stay compact too.

Whenever we have repetition this is a bad sign.

For `#asWord` construct that appears when reading from memory, you might need some additional lemmas like the lemmas that I added at the end of verification.k.

In principle I can do this refactoring, but then it will interfere with your work because you'll have to merge your current changes with refactored spec.

resources/lemmas.md
yzhang90: if sizeofWordStack(W:WS) > 0,  rule nthbyteof(chop(#asWord(W : WS) &Int MASK), 0, N) will always rewrite to 0?

gnosis/test/api-test.ini
yzhang90: I think you can first go ahead and refactor it as I'm working on other stuff now. @daejunpark Could you share your opinion before the refactoring starts?

resources/lemmas.md
denis-bogdanas: No, why? `sizeofWordStack(W:WS)` is of course always > 0. And `((MASK >>Byte #sizeWordStack(WS)) &Int 255)` just extracts the byte on the right position from MASK. The piece `&Int 255` at the end is actually redundant because W is always < 255 and upper bytes of the mask won't affect the result.

gnosis/test/api-test.ini
denis-bogdanas: Yes, @daejunpark we need your opinion on this. Do you prefer Yi's implementation or I should refactor according to my proposal?

I'll merge in the meantime. The refactoring will be a separate PR.

PR #62, opened 2018-10-27
gnosis: Added side conditions to lemmas for WordStack reading from memory

denis-bogdanas:


denis-bogdanas:
@yzhang90 Can you approve this pls so that I can push something else?

gnosis/verification.k
```diff
@@ -79,13 +79,19 @@ module VERIFICATION
...
     rule nthbyteof(WS, I, N) &Int 255 => nthbyteof(WS, I, N)
 
     rule nthbyteof(chop(MASK &Int #asWord(WS)) => chop(#asWord(WS) &Int MASK), _, _)
       requires #isConcrete(MASK)
       //version below doesn't work
       //requires notBool (getKLabel(MASK) ==K #klabel(`#asWord`))
 
+    // #asWord only keeps 32 bytes and chops the rest.
+    rule #asWord(W : WS => WS)
+      requires #sizeWordStack(W : WS) >Int 32
```
daejunpark: have you seen this case actually happens?

denis-bogdanas: Yes. I've seen an #asWord with ~192 elements, The rules for `nthbyteof(#asWord() &Int MASK)` were applying and the spec was proved, although this was not correct.

daejunpark: Would it be a potential bug of the program by any chance?

I mean, we don't know yet whether the original concrete #asWord definition is correct or not, in case that the argument is more than 32 bytes.  It is never tested in that case.

Could you elaborate a bit more in which context it happens?

denis-bogdanas: I didn't study it too much. Yi noticed that semantics of #asWord chops result to 32 byes. I added the right side condition to lemmas dealing with #asWord and got a failing spec because some #asWord had an argument longer than 32. Added this trimming lemma and all was good again.

We can possibly find the details in the jenkins log history.

denis-bogdanas: Best way to proceed: after this PR is merged, I'll delete the lemma and PR again, let's see what happens.

resources/lemmas.md
```diff
@@ -85,6 +86,21 @@ The following lemmas essentially capture the signature extraction mechanisms.
...
+    syntax Bool ::= #isRegularWordStack ( WordStack ) [function]
+ // -------------------------------------------------------
+    rule #isRegularWordStack(N : WS => WS)
+    rule #isRegularWordStack(.WordStack) => true
+
+    //Rules for #padToWidth with regular symbolic arguments.
+    //Same as for concrete #padToWidth, when WordStack is of regular form "A:B ... :.WordStack"
+    //Not clear why KEVM rules for #padToWidth were marked [concrete]. If they were general, rules below would not be necessary.
+    rule #padToWidth(N, WS) => WS
+      requires notBool #sizeWordStack(WS) <Int N andBool #isRegularWordStack(WS) ==K true
```
daejunpark: are you sure that this is correctly parsed as you intended (without extra parentheses)?

resources/lemmas.md
```diff
@@ -85,6 +86,21 @@ The following lemmas essentially capture the signature extraction mechanisms.
...
 
 ```k
+    syntax Bool ::= #isRegularWordStack ( WordStack ) [function]
+ // -------------------------------------------------------
+    rule #isRegularWordStack(N : WS => WS)
+    rule #isRegularWordStack(.WordStack) => true
+
+    //Rules for #padToWidth with regular symbolic arguments.
+    //Same as for concrete #padToWidth, when WordStack is of regular form "A:B ... :.WordStack"
+    //Not clear why KEVM rules for #padToWidth were marked [concrete]. If they were general, rules below would not be necessary.
```
daejunpark: hmm,, good point.  the reason why i added [concrete] attribute is to ensure that the rules are deterministic even in the presence of the lemma rules in lemmas.md.  but it may not be necessary.

why don't you try dropping [concrete] attribute there, and see if the tests pass? (you can do that in a separate PR though.)

gnosis/verification.k
```diff
@@ -72,22 +72,52 @@ module VERIFICATION
...
 
-    syntax WordStack ::= #encSig (Int, Int, Int) [function]
+
+    syntax WordStack ::= #bytesSegment (Int, Int, Int, Int) [function] //DATA, START, RANGE_LEN, DATA_LEN
  // ===============================================================
-    rule #encSig(SIG_R, SIG_S, SIG_V) => #enc(#bytes32(SIG_R)) ++ #enc(#bytes32(SIG_S)) ++ #getValue(#uint8(SIG_V)) : .WordStack
+    rule #bytesSegment(DATA, START, LEN, N)
+         => nthbyteof(DATA, START, N) : #bytesSegment(DATA, START +Int 1, LEN -Int 1, N)
+      requires #range ( 0 <= START < N ) andBool #range( 1 <= START +Int LEN <= N ) andBool LEN >Int 0
```
daejunpark: this could be double-edge sword.  sometimes we may want to keep the whole #bytesSegment term as it is, especially when we reason about the variable length buffer.  keep that in mind.

gnosis/verification.k
daejunpark: sounds a good plan

gnosis/verification.k
```diff
@@ -72,22 +72,52 @@ module VERIFICATION
...
+
+    rule nthbyteof(#asWord(W : WS) &Int MASK, 0, N) => W &Int (MASK >>Byte #sizeWordStack(WS))
+      requires #sizeWordStack(W : WS) <=Int 32
+
+    //decoding nthbyteof() &Int MASK
+    rule MASK &Int nthbyteof(WS, I, N) => nthbyteof(WS, I, N) &Int MASK
+      requires #isConcrete(MASK)
+
+    rule nthbyteof(_, _, _)  &Int (A => A &Int 255)
+      requires A >Int 255
```
daejunpark: i cannot see what it means. can you explain?

gnosis/verification.k
```diff
@@ -72,22 +72,52 @@ module VERIFICATION
...
-      => nthbyteof(chop(#asWord(    WS) &Int MASK), I -Int 1, N -Int 1)
-      requires I >Int 0
+    //decoding nthbyteof(#asWord())
+    rule nthbyteof(#asWord(W : WS), I       , N       )
+      => nthbyteof(#asWord(    WS), I -Int 1, N -Int 1)
+      requires I >Int 0 andBool #sizeWordStack(W : WS) <=Int 32
 
-    rule nthbyteof(chop(#asWord(W : WS) &Int MASK), 0, N) => W &Int ((MASK >>Byte #sizeWordStack(WS)) &Int 255)
+    rule nthbyteof(#asWord(W : WS), 0, N) => W
+      requires #sizeWordStack(W : WS) <=Int 32
```
daejunpark: i think it should be `#sizeWordStack(W : WS) <=Int N`

gnosis/verification.k
```diff
@@ -72,22 +72,52 @@ module VERIFICATION
 
 // Reading symbolic WordStack from memory, then extracting bytes from it. General case (stack size can be != 32)
 //========================================================================
-    rule nthbyteof(chop(#asWord(W : WS) &Int MASK), I       , N       )
-      => nthbyteof(chop(#asWord(    WS) &Int MASK), I -Int 1, N -Int 1)
-      requires I >Int 0
+    //decoding nthbyteof(#asWord())
+    rule nthbyteof(#asWord(W : WS), I       , N       )
+      => nthbyteof(#asWord(    WS), I -Int 1, N -Int 1)
+      requires I >Int 0 andBool #sizeWordStack(W : WS) <=Int 32
```
daejunpark: `#sizeWordStack(W : WS) <=Int N`

gnosis/verification.k
```diff
@@ -72,22 +72,52 @@ module VERIFICATION
...
+    rule nthbyteof(#asWord(W : WS), 0, N) => W
+      requires #sizeWordStack(W : WS) <=Int 32
+
+    //decoding nthbyteof(#asWord() &Int MASK)
+    rule nthbyteof(#asWord(W : WS) &Int MASK, I       , N       )
+      => nthbyteof(#asWord(    WS) &Int MASK, I -Int 1, N -Int 1)
+      requires I >Int 0 andBool #sizeWordStack(W : WS) <=Int 32
+
+    rule nthbyteof(#asWord(W : WS) &Int MASK, 0, N) => W &Int (MASK >>Byte #sizeWordStack(WS))
+      requires #sizeWordStack(W : WS) <=Int 32
```
daejunpark: #sizeWordStack(W : WS) <=Int N

gnosis/verification.k
```diff
@@ -72,22 +72,52 @@ module VERIFICATION
...
+      requires I >Int 0 andBool #sizeWordStack(W : WS) <=Int 32
 
-    rule nthbyteof(chop(#asWord(W : WS) &Int MASK), 0, N) => W &Int ((MASK >>Byte #sizeWordStack(WS)) &Int 255)
+    rule nthbyteof(#asWord(W : WS), 0, N) => W
+      requires #sizeWordStack(W : WS) <=Int 32
+
+    //decoding nthbyteof(#asWord() &Int MASK)
+    rule nthbyteof(#asWord(W : WS) &Int MASK, I       , N       )
+      => nthbyteof(#asWord(    WS) &Int MASK, I -Int 1, N -Int 1)
+      requires I >Int 0 andBool #sizeWordStack(W : WS) <=Int 32
```
daejunpark: #sizeWordStack(W : WS) <=Int N

resources/lemmas.md
denis-bogdanas: yes, just verified in domains.k

gnosis/verification.k
denis-bogdanas: It trims the mask to the last byte only. Because nthbyteof has only one byte. Side condition is required to avoid rewriting to itself.

gnosis/verification.k
denis-bogdanas: Good catch! Actually it should be `#sizeWordStack(W : WS) ==Int N andBool N <=Int 32` If it's strictly `#sizeWordStack(W : WS) <Int N` then wrong byte will be selected.

Will correct here and below.

resources/lemmas.md
denis-bogdanas: will do that in another PR.

gnosis/verification.k
denis-bogdanas: Here it's correct. `#sizeWordStack(W : WS) <=Int 32` is required to force #asWord trimming rule to apply.

gnosis/verification.k
denis-bogdanas: corrected

gnosis/verification.k
denis-bogdanas: Already correct.

gnosis/verification.k
daejunpark: Counter example.

```
nthbyteof(#asWord(W1 : W2 : W3 : .WordStack), 1, 4)  ==  W1
nthbyteof(#asWord(     W2 : W3 : .WordStack), 0, 3)  ==  0
```


Also, not sure it is OK without requiring `I <Int N`.

daejunpark: My suggestion:

`0 <Int I andBool I <Int N andBool N <=Int 32` and

```
andBool N <=Int #sizeWordStack(W : WS)
```
or
```
andBool N ==Int #sizeWordStack(W : WS)
```

gnosis/verification.k
daejunpark: See above.

daejunpark:
please address the rules that are still problematic.

gnosis/verification.k
denis-bogdanas: Oh, now I see the issue with this lemma. Corrected.

denis-bogdanas:
Done.

PR #64, opened 2018-10-30
Update spec for testSignatureSplit using new memory abstraction

yzhang90:


gnosis/test/api-test.ini
```diff
@@ -163,42 +163,25 @@ proxy_storage:
     andBool #rangeBytes( 32,        DOMAIN_SEPARATOR)
 
 [testSignatureSplit]
+output: _ => #encodeArgs(#uint8  (#asWord(#bufSeg(#buf(LEN, SIGS), 65 *Int POS +Int 64, 1 ))),
+                         #bytes32(#asWord(#bufSeg(#buf(LEN, SIGS), 65 *Int POS        , 32))),
+                         #bytes32(#asWord(#bufSeg(#buf(LEN, SIGS), 65 *Int POS +Int 32, 32))))
```
daejunpark: Awesome!!

gnosis/test/api-test.ini
```diff
@@ -163,42 +163,25 @@ proxy_storage:
...
 msg_sender: MSG_SENDER
-callData: #abiCallData("testSignatureSplit", #bytes(#encSig(SIG0_R, SIG0_S, SIG0_V) ++ #encSig(SIG1_R, SIG1_S, SIG1_V) ++ #encSig(SIG2_R, SIG2_S, SIG2_V)), #uint256(POS))
+callData: #abiCallData("testSignatureSplit", #bytes(#buf(LEN, SIGS)), #uint256(POS))
 callValue: 0
 proxy_storage:
     _:Map
 +requires:
     // Range
     andBool #rangeAddress(PROXY_ID)
-    andBool #rangeAddress(MSG_SENDER)
```
daejunpark: is this omission orthogonal?

resources/lemmas.md
```diff
@@ -11,6 +11,16 @@ module LEMMAS
     imports K-REFLECTION
 ```
 
+### Gas Abstraction
+```k
+    syntax Int ::= "#infinity" [function]
+ // -------------------------------------
+    rule #infinity -Int _ => #infinity
+    rule #infinity >=Int _ => true
+    rule #infinity <Int _ => false
```
daejunpark: could you put some comments here, saying that it is temporary?

resources/lemmas.md
```diff
@@ -207,6 +213,12 @@ These rules are specific to reasoning about EVM programs.
     rule (I1 -Int I2) +Int I3 => I1 -Int (I2 -Int I3) when #isConcrete(I2) andBool #isConcrete(I3)
     rule (I1 -Int I2) -Int I3 => I1 -Int (I2 +Int I3) when #isConcrete(I2) andBool #isConcrete(I3)
 
+    rule I1 +Int (I2 +Int I3) => I2 +Int (I1 +Int I3) when #isConcrete(I1) andBool #isConcrete(I3)
+    rule I1 +Int (I2 +Int I3) => I3 +Int (I1 +Int I2) when #isConcrete(I1) andBool #isConcrete(I2)
+
+    rule (I1 +Int I2) +Int (I3 -Int I2) => I1 +Int I3
+    rule (I1 +Int I2) -Int (I1 +Int I3) => I2 -Int I3
```
daejunpark: all of these four rules are related to gas cost calculation, right?

resources/lemmas.md
```diff
@@ -342,10 +321,8 @@ These lemmas abstract some properties about `#sizeWordStack`:
 
 ```k
     rule #sizeWordStack ( _ , _ ) >=Int 0 => true [smt-lemma]
-    rule #sizeWordStack ( WS , N:Int )
-      => #sizeWordStack ( WS , 0 ) +Int N
-      requires N =/=K 0
-      [lemma]
+
+    rule WS ++ .WordStack => WS
```
daejunpark: i'm not sure if this is safe to add.
note that #take and #drop actually does the opposite direction rewriting, from BUF => BUF ++ .WordStack.
do you mean here `notBool #isBuf(WS)`?

resources/lemmas.md
```diff
@@ -218,41 +230,8 @@ These rules are specific to reasoning about EVM programs.
...
-       andBool 0 <=Int B25 andBool B25 <Int 256
-       andBool 0 <=Int B26 andBool B26 <Int 256
-       andBool 0 <=Int B27 andBool B27 <Int 256
-       andBool 0 <=Int B28 andBool B28 <Int 256
-       andBool 0 <=Int B29 andBool B29 <Int 256
-       andBool 0 <=Int B30 andBool B30 <Int 256
-       andBool 0 <=Int B31 andBool B31 <Int 256
-       andBool 0 <=Int B32 andBool B32 <Int 256
+    rule #asWord(#bufSeg(BUF, START, WIDTH)) &Int 255 => #asWord(#bufSeg(BUF, START +Int WIDTH -Int 1, 1))  requires WIDTH >=Int 1
+    rule 255 &Int #asWord(#bufSeg(BUF, START, WIDTH)) => #asWord(#bufSeg(BUF, START +Int WIDTH -Int 1, 1))  requires WIDTH >=Int 1
```
daejunpark: don't we need `WIDTH <=Int 32` as well to be safe?

resources/lemmas.md
```diff
@@ -72,6 +82,13 @@ They capture the essential mechanisms used by the two instructions: splitting a
                 : nthbyteof(V, 31, 32)
                 : .WordStack ) => V
       requires 0 <=Int V andBool V <Int pow256
+
+    rule 0 <=Int #asWord(#bufSeg(_, _, _))                 => true
+    rule         #asWord(#bufSeg(_, _, WIDTH)) <Int pow256 => true  requires WIDTH <=Int 32
+
+    rule         #asWord(#bufSeg(_, _, 1)) <=Int 255 => true
```
daejunpark: i guess having the side-condition of `WIDTH ==Int 1` is more applicable than directly matching `1`.

resources/lemmas.md
```diff
@@ -72,6 +82,13 @@ They capture the essential mechanisms used by the two instructions: splitting a
...
+
+    rule #buf(32, #asWord(#bufSeg(BUF, START, WIDTH))) => #bufSeg(BUF, START +Int WIDTH -Int 32, 32)  requires WIDTH >=Int 32
```
daejunpark: i'm surprised here. is this really happens in the bytecode? could you elaborate the scenario?

gnosis/test/api-test.ini
yzhang90: remove by mistake. added it back

resources/lemmas.md
yzhang90: done

resources/lemmas.md
yzhang90: No. To simply the representation of the address for memory lookup.

resources/lemmas.md
yzhang90: Update the evm-semantics to be compatible with the rule.

resources/lemmas.md
yzhang90: This is a more general rule. The actual instance used in proving is when WIDTH ==Int 32. rhs is the actual representation that appears during the execution. lhs is the result of `#bytes32(#asWord(#bufSeg(#buf(LEN, SIGS), 65 *Int POS        , 32)))` (which becomes `#buf(32, #asWord(#bufSeg(#buf(LEN, SIGS), 65 *Int POS        , 32)))`

resources/lemmas.md
```diff
@@ -72,6 +83,13 @@ They capture the essential mechanisms used by the two instructions: splitting a
                 : nthbyteof(V, 31, 32)
                 : .WordStack ) => V
       requires 0 <=Int V andBool V <Int pow256
+
+    rule 0 <=Int #asWord(#bufSeg(_, _, _))                 => true
+    rule         #asWord(#bufSeg(_, _, WIDTH)) <Int pow256 => true  requires WIDTH <=Int 32
```
daejunpark: if we want to allow #asWord take more than 32 bytes, the `requires` is not needed i think.

yzhang90: agreed

resources/lemmas.md
daejunpark: to be more general, i think we need to define the case when WIDTH < 32?

resources/lemmas.md
daejunpark: if we allow WIDTH > 32, then i agree.

PR #65, opened 2018-10-30
add origStorage cell

dwightguth:
To be merged together with kframework/evm-semantics#267


dwightguth:
I'm reasonably certain this is failing because it's using the wrong version of the evm semantics. Once the evm pr is merged, I will update this with an update to the submodule.

dwightguth:
Can I get a review on this? The tests are now passing.

PR #69, opened 2018-11-07
testEcrecover full specification

denis-bogdanas:
Gnosis: test spec for ecrecover, full specification. Non-empty and empty case. Uses symbolic rules for ECREC in KEVM-gnosis and latest formula simplification improvements in K-gnosis.

https://github.com/kframework/evm-semantics/blob/9a409babcd9b77a0f9a30f52350e4c5d46e6b086/symbolic.md

https://github.com/kframework/k/pull/191


gnosis/verification.k
```diff
@@ -120,4 +120,14 @@ module VERIFICATION
...
 
+// Range for #symEcrec
+//========================================================================
+    rule 0 <=Int  #symEcrec(DATA)             => true
+
+    //that's because the result in concrete semantics is trimmed to Address range.
+    rule          #symEcrec(DATA) <Int pow160 => true
+
+    //case 0 is never wrapped into #symEcrec(), corresponds to #ecrecEmpty(DATA) == true
+    rule 0 =/=Int #symEcrec(DATA)             => true
```
daejunpark: maybe we also need this type of thing for the normal address validity condition:
```
andBool MASTER_COPY_ID >Int 0 andBool (notBool MASTER_COPY_ID in #precompiledAccounts(BYZANTIUM))
```

gnosis/test/api-test.ini
```diff
@@ -199,9 +199,50 @@ proxy_storage:
...
+accounts:
+    <account>
+      ; precompiled account for ECCREC
+      <acctID> 1 </acctID>
+      <balance> 0 </balance>
+      <code> .WordStack </code>
+      <storage> .Map </storage>
+      <origStorage> .Map </origStorage>
+      <nonce> 0 </nonce>
+    </account>
```
daejunpark: Is this account cell general enough, or just to mock it for now?

daejunpark:
Lgtm overall

gnosis/test/api-test.ini
denis-bogdanas: This is a precompiled account, so it should be an inactive account I guess. But if I leave it inactive I get issues somewhere else, don't remember where exactly. Some symbolic terms cannot be evaluated. So for the purpose of simulating a precompiled account this is general enough.

gnosis/verification.k
denis-bogdanas: I would add this condition only if I get issues. So far it's not needed.

denis-bogdanas:
can you approve then?

gnosis/test/api-test.ini
daejunpark: I think we can revisit this later to make sure nothing is missed here.

PR #70, opened 2018-11-16
GNO Token formal verification

hacker-DOM:
I claim to have formally verified the GNO token. The code is available [here](https://etherscan.io/address/0x6810e776880c02933d47db1b9fc05908e5386b96#code).

I made only minute changes to EVM-ERC20. I renamed the storage variables, because they have a different name in GNO. I removed the overflow situations, because GNO was created with a supply of 10 M, and I provide a proof that the supply is invariant. I also verified cases when callValue > 0. I had to add a lemma, the proofs don't run without it.

Do you think it would be possible to approve this, say by Friday 23/11? See slack for more info

### DIFF
--- a/resources/lemmas.md
+++ b/resources/lemmas.md
@@ -337,6 +337,7 @@ The other rules are similar.
 
 ```k
     rule chop ( W0:Int +Int W1:Int ) -Word W1:Int => chop ( W0 )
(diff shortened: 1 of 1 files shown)


denis-bogdanas:
jenkins: test this please

PR #71, opened 2018-11-16
Proxied token formal verification

dteiml:
I claim to have formally verified a proxy token with a ERC20 StandardToken as its masterCopy. I can provide a README if you suggest, but everything is already present in [spec-tmpl.k](./proxied-token/spec-tmpl.k) and [proxied-token-spec.ini](./proxied-token/proxied-token-spec.ini). Let me know here if you have any questions!

### DIFF
--- a/resources/lemmas.md
+++ b/resources/lemmas.md
@@ -3,12 +3,9 @@ Verification Lemmas
 
 ```k
 requires "evm.k"
-requires "edsl.k"
 
 module LEMMAS
     imports EVM
-    imports EDSL
-    imports K-REFLECTION
 ```
 
 ### Memory Abstraction
@@ -35,6 +32,9 @@ The following lemmas are used for symbolic reasoning about `MLOAD` and `MSTORE` 
 They capture the essential mechanisms used by the two instructions: splitting a word into the byte-array and merging it back to the word.
 
 ```k
+    rule 0 <=Int nthbyteof(V, I, N)          => true
+    rule         nthbyteof(V, I, N) <Int 256 => true
+
     rule #asWord( nthbyteof(V,  0, 32)
                 : nthbyteof(V,  1, 32)
                 : nthbyteof(V,  2, 32)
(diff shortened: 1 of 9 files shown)


dteiml:
I had to add 3 new lemmas (one is shared with #70). I looked deeply into verifying they are all correct and that they are all necessary.

resources/lemmas.md
```diff
@@ -333,6 +333,29 @@ The other rules are similar.
...
+
+```k
+    rule chop ( W0:Int +Int W1:Int ) -Word W1:Int => chop ( W0 )
+    requires #rangeUInt(256, W0) andBool #rangeUInt(256, W1)
+```
+
+### `modInt` Reduction
+
+```k
+    rule X:Int modInt 1461501637330902918203684832716283019655932542976 => X
```
denis-bogdanas: There is a more general lemma for modInt in gnosis/verification.k and other versions of verification.k. Please use that version.

resources/lemmas.md
```diff
@@ -333,6 +333,29 @@ The other rules are similar.
...
+        requires #rangeAddress(X)
+```
+
+### `&Int` Reduction
+
+An Ethereum address X has an address space of [0, 2^160). Hence if we do (2^256 - 1) - (2^160 - 1) &Int X, the result will always be 0.
+
+```k
+    rule 115792089237316195423570985007226406215939081747436879206741300988257197096960 &Int X:Int => 0
```
denis-bogdanas: Can you write a generalized lemma, similar to rule that starts with `rule MASK &Int N => N` here? https://github.com/runtimeverification/verified-smart-contracts/pull/71/files

denis-bogdanas:
Yes, please write a short readme with link to the source code and mention of the commit/tag that you used.

resources/lemmas.md
dteiml: Done

resources/lemmas.md
dteiml: I'm not sure I'm able to generalize this. Graphically the rule says that f..f (24 times) ++ 0..0 (40 times) will AND with any Ethereum address to 0. (Reason is that addresses have at most 40 hex digits.) I could maybe generalize it by extending it to _any_ number N with N = 0 mod 2^160, but that might introduce new complexity and irreducible terms.

PR #73, opened 2018-11-19
add specs for testKeccak

yzhang90:


denis-bogdanas:
@yzhang90 Can you generate bytecode from this version? https://github.com/denis-bogdanas/safe-contracts/tree/rv
Branch `rv`. It contains the latest fix on their end, and 2 functions made public to facilitate verification.

PR #75, opened 2018-11-20
Proxied token formal verification #71

daejunpark:
reissue #71

### DIFF
--- a/proxied-token/proxied-token-spec.ini
+++ b/proxied-token/proxied-token-spec.ini
@@ -1,0 +1,295 @@
+[root]
+k: #execute => #halt
+code: {PROXY_CODE}
+
+; This are our variables:
+; Default statusCode is EVMC_REVERT
+statusCode: _ => EVMC_REVERT
+output: _ => _
+callData:
+log: _
+refund: _ => _
+msg_sender: RANDOM_ID
+callValue: 0
+storage:
+    #hashedLocation({COMPILER}, {_MASTER_COPY}, .IntList) |-> MASTER_COPY_ID
+requires:
+
+; Proxy pattern functions
+
+[implementation]
+callData: #abiCallData("implementation", .TypedArgs)
+
+[implementation-success]
+statusCode: _ => EVMC_SUCCESS
+output: _ => #asByteStackInWidth(MASTER_COPY_ID, 32)
+
+[implementation-failure]
+; non-zero call value
+callValue: CALL_VALUE
++requires:
+    andBool #rangeUInt(256, CALL_VALUE)
+    andBool 0  <Int CALL_VALUE
+
+[proxyType]
+callData: #abiCallData("proxyType", .TypedArgs)
+
+[proxyType-success]
+statusCode: _ => EVMC_SUCCESS
+output: _ => #asByteStackInWidth(2, 32)
+
+[proxyType-failure]
+; non-zero call value
+callValue: CALL_VALUE
++requires:
+    andBool #rangeUInt(256, CALL_VALUE)
+    andBool 0  <Int CALL_VALUE
+
+[changeMasterCopy]
+callData: #abiCallData("changeMasterCopy", #address(NEW_MASTER_COPY_ID))
++storage:
+    #hashedLocation({COMPILER}, {_CREATOR}, .IntList) |-> CREATOR_ID
++requires:
+    andBool #rangeAddress(NEW_MASTER_COPY_ID)
+
+[changeMasterCopy-success]
+statusCode: _ => EVMC_SUCCESS
+msg_sender: CREATOR_ID
+storage:
+    #hashedLocation({COMPILER}, {_MASTER_COPY}, .IntList) |-> (MASTER_COPY_ID => NEW_MASTER_COPY_ID)
+    #hashedLocation({COMPILER}, {_CREATOR}, .IntList) |-> CREATOR_ID
++requires:
+    andBool 0  <Int  NEW_MASTER_COPY_ID 
+
+[changeMasterCopy-failure]
+
+[changeMasterCopy-failure-1]
+; masterCopy with address 0 supplied
+msg_sender: CREATOR_ID
++requires:
+    andBool 0 ==Int NEW_MASTER_COPY_ID
+
+[changeMasterCopy-failure-2]
+; function called by address different from creator
++requires:
+    andBool #rangeAddress(NEW_MASTER_COPY_ID)
+    andBool 0  <Int  NEW_MASTER_COPY_ID
+
+[changeMasterCopy-failure-3]
+; non-zero call value
+callValue: CALL_VALUE
+msg_sender: CREATOR_ID
++requires:
+    andBool #rangeUInt(256, CALL_VALUE)
+    andBool 0  <Int  CALL_VALUE
+    andBool #rangeAddress(NEW_MASTER_COPY_ID)
+    andBool 0  <Int  NEW_MASTER_COPY_ID
+
+; ERC20 functions
+
+[totalSupply]
+callData: #abiCallData("totalSupply", .TypedArgs)
+
+[totalSupply-success]
+statusCode: _ => EVMC_SUCCESS
+output: _ => #asByteStackInWidth(TOTAL, 32)
++storage:
+    #hashedLocation({COMPILER}, {_TOTALTOKENS}, .IntList) |-> TOTAL
++requires:
+    andBool #rangeUInt(256, TOTAL)
+
+[totalSupply-failure]
+callValue: CALL_VALUE
++requires:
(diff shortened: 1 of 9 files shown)


proxied-token/verification.k
```diff
@@ -0,0 +1,18 @@
...
+    rule X modInt Y => X
+      requires 0 <=Int X andBool X <Int Y
+
+    rule chop ( W0:Int +Int W1:Int ) -Word W1:Int => chop ( W0 )
+        requires #rangeUInt(256, W0) andBool #rangeUInt(256, W1)
+
+    // An Ethereum address X has an address space of [0, 2^160). Hence if we do (2^256 - 1) - (2^160 - 1) &Int X, the result will always be 0.
+
+    rule 115792089237316195423570985007226406215939081747436879206741300988257197096960 &Int X:Int => 0
+        requires #rangeAddress(X)
```
denis-bogdanas: Please add a comment explaining the meaning of this number - e.g. its hex value.
Also, for all other lemmas please explain in comments where are they needed to the extent you know.

proxied-token/proxied-token-spec.ini
```diff
@@ -0,0 +1,295 @@
...
+[changeMasterCopy-failure-1]
+; masterCopy with address 0 supplied
+msg_sender: CREATOR_ID
++requires:
+    andBool 0 ==Int NEW_MASTER_COPY_ID
+
+[changeMasterCopy-failure-2]
+; function called by address different from creator
++requires:
+    andBool #rangeAddress(NEW_MASTER_COPY_ID)
```
denis-bogdanas: duplicate range predicate.
Also this spec is ambiguous to read. Please explicitly add msg_sender value here, even if it is inherited.

proxied-token/proxied-token-spec.ini
```diff
@@ -0,0 +1,295 @@
...
+[changeMasterCopy-success]
+statusCode: _ => EVMC_SUCCESS
+msg_sender: CREATOR_ID
+storage:
+    #hashedLocation({COMPILER}, {_MASTER_COPY}, .IntList) |-> (MASTER_COPY_ID => NEW_MASTER_COPY_ID)
+    #hashedLocation({COMPILER}, {_CREATOR}, .IntList) |-> CREATOR_ID
++requires:
+    andBool 0  <Int  NEW_MASTER_COPY_ID 
+
+[changeMasterCopy-failure]
```
denis-bogdanas: empty intermediate specs like this have no effect. Please delete.

proxied-token/proxied-token-spec.ini
```diff
@@ -0,0 +1,295 @@
...
+[implementation-success]
+statusCode: _ => EVMC_SUCCESS
+output: _ => #asByteStackInWidth(MASTER_COPY_ID, 32)
+
+[implementation-failure]
+; non-zero call value
+callValue: CALL_VALUE
++requires:
+    andBool #rangeUInt(256, CALL_VALUE)
+    andBool 0  <Int CALL_VALUE
```
denis-bogdanas: This sort of failure specs also look ambiguous. Please add status code specification for all.

proxied-token/proxied-token-spec.ini
```diff
@@ -0,0 +1,295 @@
...
++requires:
+    andBool RANDOM_ID ==Int TO_ID
+    andBool VALUE <=Int BAL_FROM
+    andBool BAL_FROM +Int VALUE <Int (2 ^Int 256)
+
+[transfer-failure]
+output: _ => #asByteStackInWidth(0, 32)
+[transfer-failure-1]
++storage:
+    #hashedLocation({COMPILER}, {_BALANCES}, RANDOM_ID) |-> (BAL_FROM => _)
```
denis-bogdanas: is BAL_FROM changing in case of failure??

proxied-token/proxied-token-spec.ini
```diff
@@ -0,0 +1,295 @@
...
+    #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)     |->  BAL_TO
++requires:
+    andBool ( VALUE >Int BAL_FROM
+     orBool   BAL_TO +Int VALUE >=Int (2 ^Int 256) )
+
+[transfer-failure-2]
```
denis-bogdanas: how is this a failure?

proxied-token/proxied-token-spec.ini
```diff
@@ -0,0 +1,295 @@
...
+
+[transfer-failure]
+output: _ => #asByteStackInWidth(0, 32)
+[transfer-failure-1]
++storage:
+    #hashedLocation({COMPILER}, {_BALANCES}, RANDOM_ID) |-> (BAL_FROM => _)
+    #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)     |->  BAL_TO
++requires:
+    andBool ( VALUE >Int BAL_FROM
+     orBool   BAL_TO +Int VALUE >=Int (2 ^Int 256) )
```
denis-bogdanas: Use pow256 here.

denis-bogdanas: Also, looks like status is EVMC_SUCCESS, inherited from `[transfer]`. In this case the token does not comply with ERC20. Please correct me if I'm wrong. Same for other failure rules.

proxied-token/proxied-token-spec.ini
```diff
@@ -0,0 +1,295 @@
...
+[transfer-success-2]
++storage:
+    #hashedLocation({COMPILER}, {_BALANCES}, RANDOM_ID) |-> BAL_FROM
++requires:
+    andBool RANDOM_ID ==Int TO_ID
+    andBool VALUE <=Int BAL_FROM
+    andBool BAL_FROM +Int VALUE <Int (2 ^Int 256)
+
+[transfer-failure]
+output: _ => #asByteStackInWidth(0, 32)
```
denis-bogdanas: add a comment: ; return false

proxied-token/proxied-token-spec.ini
```diff
@@ -0,0 +1,295 @@
...
+    #hashedLocation({COMPILER}, {_ALLOWANCES}, RANDOM_ID SPENDER) |-> (_:Int => VALUE)
+
+[approve-failure]
+callValue: CALL_VALUE
++requires:
+    andBool #rangeUInt(256, CALL_VALUE)
+    andBool 0  <Int  CALL_VALUE
+
+[transfer]
+statusCode: _ => EVMC_SUCCESS
```
denis-bogdanas: Explain in a comment why return status is success even for most failure rules - because return value false is used instead to denote failure.

This is a difference from behavior recommended in the standard and should be noted in the readme. Please describe precisely the failure cases where functions return false instead of THROW.

proxied-token/verification.k
dteiml: First point - done. Second point - honestly I don't remember, but I'll keep a record of this next time.

proxied-token/proxied-token-spec.ini
dteiml: done

proxied-token/proxied-token-spec.ini
dteiml: done

proxied-token/proxied-token-spec.ini
dteiml: done

proxied-token/proxied-token-spec.ini
dteiml: this was an actual bug. good catch!

proxied-token/proxied-token-spec.ini
dteiml: both transfer-failure-1 and transfer-failure-2 are failures because the math overflows, so the arithmetic checks would return false (see first few lines of transfer fn in deployed.sol)

proxied-token/proxied-token-spec.ini
dteiml: pow256 - done
It does comply with ERC20- the definition of compliance is that it implements all MUST clauses. Throwing is a SHOULD clause.

proxied-token/proxied-token-spec.ini
dteiml: done

proxied-token/proxied-token-spec.ini
dteiml: done

denis-bogdanas:
Also please rebase both PRs on master. Gnosis branch is work in progress.

dteiml:
> Also please rebase both PRs on master. Gnosis branch is work in progress.

It is already rebased on master. My first commit `Finished proxied token` e1975 has as parent `testEcrecover full specification (#69)` 995d4, latest master commit

PR #81, opened 2018-11-26
erc20,Makefile: Added support for Constantinople.

denisadiaconescu:


denis-bogdanas:
@denisadiaconescu You are using a version of KEVM that is not compatible with master. That's why the test fails. Please rollback the changes in .kevm.rev.

We'll use latest kevm after my changes get merged. Hope we won't have this sort of issues in future.

daejunpark:
any reason of having two versions of the specs, instead of having only constantinople one?

denisadiaconescu:
I wasn't sure if we were going to use the byzantium version as well. If you think we drop the byzantium version, we can make the changes and keep just the constantinople version.

daejunpark:
Yes, I think we don't need to maintain both. Please drop the old version.

denisadiaconescu:
I dropped the byzantium versions in erc20 and rebased master. Should we transition the new `erc20/gno` to constantinople as well?

erc20/ds-token/ds-token-erc20-spec.ini
```diff
@@ -1,265 +1,301 @@
...
+gasCap: 100000
```
daejunpark: why the diff shows entire file is changed? did you change the file format  dos <-> unix  stuff?

erc20/spec-tmpl.k
```diff
@@ -72,10 +72,8 @@
             <acctID> ACCT_ID </acctID>
             <balance> _ </balance>
             <code> #parseByteStack({CODE}) </code>
-            <storage>
-              {STORAGE}
-            </storage>
-            <origStorage> _ </origStorage>
+            <storage>{STORAGE} </storage>
```
daejunpark: nitpick: add space between <storage> and {STORAGE}.

daejunpark:
looks good. could you check the K version? i just merged another PR, and now the `.k.rev` is conflicting. also please address my minor comments.

btw, the jenkins test will fail, as i'm doing the makefile splitting, which requires to modify the jenkins command. i'll fix it.

erc20/ds-token/ds-token-erc20-spec.ini
denisadiaconescu: Looks it was indeed CRLF instead of LF. I fixed it.

daejunpark:
ok. i'll first merge my stuff, and rebase this. we'll see if jenkins passes.

daejunpark:
I rebased, but the jenkins failed. Could you take a look at?

denis-bogdanas:
jenkins: test this please

denis-bogdanas:
@denisadiaconescu Possibly Jenkins configuration issue - looks like it deserializes data saved by earlier versions of kprove. Make a change to trigger the build. If it passes, @daejunpark maybe you can take a look? Dwight had a similar problem recently.

PR #83, opened 2018-11-27
split Makefile for gnosis

daejunpark:


daejunpark:
this is a step toward to split Makefile for each contract directory that can have customized version of k and kevm.

PR #90, opened 2018-11-30
fix ambiguous lemma

daejunpark:


daejunpark:
@denis-bogdanas Yes, Makefile can specify custom locations. See this:
https://github.com/runtimeverification/verified-smart-contracts/blob/fea6e4e8dc3192667201168dd4767574672d0b93/gnosis/test/Makefile#L3-L4

I agree we will need to avoid duplication and share as much as possible, which should be doable.

yzhang90:
looks good to me. Hope we could use the same spec-tmpl.k and verification.k soon

PR #92, opened 2018-12-03
Specs for checkSignatures case 3, threshold == 2 #91

daejunpark:
#91


denis-bogdanas:
@daejunpark Who should review?

k-test/smt-prelude.txt
```diff
@@ -0,0 +1,2 @@
+(set-option :auto-config false)
+(set-option :smt.mbqi true)
```
daejunpark: file extension should be `.smt2`.

resources/kprove.mak
```diff
@@ -57,6 +57,9 @@ TANGLER:=$(PANDOC_TANGLE_SUBMODULE)/tangle.lua
 LUA_PATH:=$(PANDOC_TANGLE_SUBMODULE)/?.lua;;
 export LUA_PATH
 
+K_OPTS:=-Xmx12g
+export K_OPTS
+
```
daejunpark: would it work for a machine that doesn't have enough memory?

.build/.k.rev
```diff
@@ -1 +1 @@
-b65568ade007d54dd21667f4f56f2d7471bb6596
+4a9364d1b683d31cd8029cffc9217c7b230b657c
```
daejunpark: i think we need to hold this K change until we all agree with the variable abstraction.

k-test/k-test-spec.ini
```diff
@@ -0,0 +1,56 @@
...
+[ecrecT-false-constraint]
+callData: #abiCallData("ecrecConstraint", #address(A1), #bytes32(P1), #uint8(A2V), #bytes32(P3), #bytes32(P4), #address(A3))
+
+; Passes because B is RHS-only and is existentially quantified in Z3 query.
+; Requires smt prelude option: "(set-option :smt.mbqi true)"
+[rhsOnlyVar]
+k: hash1(A:Int) => hash1(B)
+statusCode: EVMC_SUCCESS
+callData: _
+output: _
```
daejunpark: which gnosis spec requires the reasoning exercised here?

k-test/k-test-spec.ini
```diff
@@ -0,0 +1,56 @@
...
+; SPEC fails on old K. Requires abstraction of #symEcrec into fresh var in implication Z3 query.
+; #symEcrec must not have [smtlib], otherwise test is pointless.
+; function ecrecImplication(address a1, bytes32 p1, uint8 a2v, bytes32 p3, bytes32 p4, address a3)
+[ecrecT-false-implication]
+callData: #abiCallData("ecrecImplication", #address(A1), #bytes32(P1), #uint8(A2V), #bytes32(P3), #bytes32(P4), #address(A3))
+
+; SPEC fails on old K. Requires abstraction of #symEcrec into fresh var in constraint Z3 query.
+; function ecrecConstraint(address a1, bytes32 p1, uint8 a2v, bytes32 p3, bytes32 p4, address a3)
+[ecrecT-false-constraint]
+callData: #abiCallData("ecrecConstraint", #address(A1), #bytes32(P1), #uint8(A2V), #bytes32(P3), #bytes32(P4), #address(A3))
```
daejunpark: these two specs will be proved by the old K if we put the smtlib attribute, right?

daejunpark:
looks good overall except some questions.

is it possible to prove this spec without using the variable abstraction feature, by adding the smtlib attribute?  since it will take some time to merge the variable abstraction stuff, i think it would be better to merge without the k update, if possible.

also, you're no longer working on this direction (spec with concrete size of buffers), right?

daejunpark:
https://app.codeship.com/projects/29b2c070-d6a6-0136-919e-029fe3d67cd9/builds/38804454?pipeline=ccf34eb5-ab09-41d1-8c49-dde9952a2143

resources/kprove.mak
daejunpark: @denis-bogdanas please use `?=` instead of `:=`, so that we can adjust when we run `make`.

that is:
```
K_OPTS?=-Xmx12g
```
instead of:
```
K_OPTS:=-Xmx12g
```

denis-bogdanas: It possibly would, but any error in specs leads to extra paths and that leads to OOMError and hard to debug issue.

Corrected K_OPTS.

gnosis/gnosis-spec.ini
```diff
@@ -431,10 +431,12 @@ output: _ => #encodeArgs( #bool(0) )
     andBool THRESHOLD >Int NR_SIG
 
 ; case SIG_V > 26, 3rd branch in checkSignatures for loop.
-[checkSignatures-threshold-1-sigv-2]
+; THRESHOLD == 1, NR_SIG == 1 or 2
+[checkSignatures-thres-1-sig-v2]
+k: ( split(NR_SIG ==Int 1) ~> split(NR_SIG ==Int 2) => . ) ~> (#execute => #halt) ~> _
```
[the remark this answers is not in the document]
denis-bogdanas: We cannot, you are right. We would need another spec that is same except split part, [trusted].

k-test/k-test-spec.ini
denis-bogdanas: None of them, just an interesting peculiarity I discovered while trying to generate a query with existential variables. Worth putting into the test suite. When existential vars are used we need `(set-option :smt.mbqi true)`.

k-test/k-test-spec.ini
denis-bogdanas: Good question. that actually won't be enough. The argument of #symEcrec is WordStack, so we would have to also mark `Int ":" WordStack` with smtlib. And then we would have very long queries with little useful information, which will likely timeout.

This is actually a good example when we cannot use smtlib and need improved term abstraction.

.build/.k.rev
denis-bogdanas: New K is actually necessary as I explained below. The merge can wait.

denis-bogdanas:
1. See my answer above regarding smtlib.
2. Yes I switched to migrating existing specs on new memory model.

k-test/k-test-spec.ini
denis-bogdanas: PR #94 opened: experiment with old K + smtlib attributes.

resources/kprove.mak
```diff
@@ -82,6 +85,7 @@ $(K_REPO_DIR):
 $(KEVM_REPO_DIR):
 	git clone $(KEVM_REPO_URL) $(KEVM_REPO_DIR)
 	cd $(KEVM_REPO_DIR) \
+	    && git clean -fx \
```
daejunpark: @denis-bogdanas does this have an effect? I mean, this recipe is executed only when $(KEVM_REPO_DIR) does NOT exist.

daejunpark: Is this for your local copy, or for jenkins? I thought jenkins starts with a fresh workspace.

denis-bogdanas: It doesn't. We sometimes get deserialization exceptions. Especially now that we use different versions of KEVM on different branches.

PR #93, opened 2018-12-05
TEST: gnosis-imap

denis-bogdanas:
1. Gnosis project: converted all specs to IMap model. Byte array length unbounded. No other abstractions.

2. Cleaned up lemmas and fixed all non-determinisms.
3. Removed unused/duplicate files.


denis-bogdanas:
jenkins: test this please

gnosis/test/verification.k
```diff
@@ -15,10 +15,12 @@ module VERIFICATION
...
     rule #infinity -Int _ => #infinity
-    rule #infinity >=Int _ => true
-    rule _ <=Int #infinity => true
-    rule #infinity <Int _ => false
+    rule A <=Int #infinity => true
+    	requires #getKLabelString(A) =/=String "#infinity_VERIFICATION"
+    rule #infinity <Int A => false
+			requires #getKLabelString(A) =/=String "#infinity_VERIFICATION"
 
+		rule minInt(#infinity, #infinity) => #infinity
```
daejunpark: @denis-bogdanas this is bothering.  is it because you use tab instead of space?

gnosis/test/verification.k
```diff
@@ -117,14 +119,13 @@ module VERIFICATION
...
 
     // Arithmetic
-    rule I1 +Int (I2 +Int I3) => I2 +Int (I1 +Int I3) when #isConcrete(I1) andBool #isConcrete(I3)
-    rule I1 +Int (I2 +Int I3) => I3 +Int (I1 +Int I2) when #isConcrete(I1) andBool #isConcrete(I2)
-
-    rule (I1 +Int I2) +Int (I3 -Int I2) => I1 +Int I3
-    rule (I1 +Int I2) -Int (I1 +Int I3) => I2 -Int I3
+    rule (A +Int I2) +Int (I3 -Int A) => I2 +Int I3
+    rule (A +Int I2) -Int (A +Int I3) => I2 -Int I3
+		rule (A +Int I2) -Int (I3 +Int A) => I2 -Int I3
```
daejunpark: @denis-bogdanas could you explain what happens here?

resources/lemmas.md
```diff
@@ -184,12 +183,19 @@ The rules are applied only when the side-conditions are met.
...
+    rule 0 -Int (0 -Int X) => X
 
-    rule I1 &Int (I2 &Int I3) => (I1 &Int I2) &Int I3 when #isConcrete(I1) andBool #isConcrete(I2)
+    rule (A +Int I2) +Int I3 => A +Int (I2 +Int I3) when notBool #isConcrete(A) andBool #isConcrete(I2) andBool #isConcrete(I3)
+    rule I1 +Int (B +Int I3) => B +Int (I1 +Int I3) when #isConcrete(I1) andBool notBool #isConcrete(B) andBool #isConcrete(I3)    
+
+    rule I  +Int (0 -Int C) => (0 -Int C) +Int I when 				#isConcrete(I) andBool notBool #isConcrete(C)
+    rule A  +Int (0 -Int C) =>  A -Int C 				 when notBool #isConcrete(A) andBool notBool #isConcrete(C)
+
+    rule I1 &Int (I2 &Int C) => (I1 &Int I2) &Int C when #isConcrete(I1) andBool #isConcrete(I2) andBool notBool #isConcrete(C)
```
daejunpark: @denis-bogdanas i cannot see why it is needed to change +Int to -Int and change it back later.

gnosis/verification.k
```diff
@@ -43,10 +43,11 @@ module VERIFICATION
 
     // select/store key equality
 
-    rule hash2(_,_) ==K I => false
-      requires #isConcrete(I)
+    rule hash2(A,B) ==K I => false
+      requires notBool #isConcrete(hash2(A,B)) andBool #isConcrete(I)
+      				 andBool I <Int 1000000000 //random small-ish number bigger than any fixed storage location
```
daejunpark: not sure if 1000000000 is small enough. could you make it tighter?

gnosis/generated/solidity compilation instructions.txt
```diff
@@ -14,3 +14,4 @@ Hex files on this repo were cleared up. They only contain the fragment that was
 
 To generate .evm files:
     Use --asm option instead of --bin-runtime
+
```
daejunpark: you can just click the "Rebuild" button in Jenkins.

daejunpark:
@denis-bogdanas the changes made to the lemma rules are not trivial.  could you please elaborate them?  for example, i don't see why it is necessary to change -Int between +Int.  also, not sure if it is possible to orient the symbolic terms in the arithmetic expressions: i mean, what if we have nested arithmetic terms?

gnosis/test/verification.k
denis-bogdanas: fixed formatting in last forced push.

gnosis/test/verification.k
```diff
@@ -41,13 +43,14 @@ module VERIFICATION
 
     // select/store key equality
 
-    rule hash2(_,_) ==Int I => false
-      requires #isConcrete(I)
+    rule hash2(A,B) ==K I => false
+      requires         notBool #isConcrete(hash2(A,B)) andBool #isConcrete(I)
+               andBool I <Int 1000000000 //random small-ish number bigger than any fixed storage location
```
daejunpark: i don't think you need `notBool #isConcrete(hash2(A,B))`, which is always true.

gnosis/test/verification.k
denis-bogdanas: Trying to separate symbolic terms of +/- from concrete terms, so that concrete terms all add up at the same side of +Int and no matter which path is taken by some other functions, final expression always looks the same.

Principles are explained in comments: move symbolic terms always to the left, convert -Int A into +Int (0 -Int A) to have always +Int, except when LHS is 0 or another symbolic term. This drastically reduces the number of particular cases elsewhere, like the +Int simplification expressions in verification.k

resources/lemmas.md
denis-bogdanas: It reduces the number of particular cases. Otherwise we would need way more rules to achieve same generality. We would need rules for both (A-I1)+I2, (A+I1)-I2 and many others.

gnosis/verification.k
denis-bogdanas: Why not:) Just pick a value. But you got the point of this side condition.
10^9 ~=2^30 is still tens of orders of magnitude less than 2^256 > 10^75, are you afraid it's so big it could lead to collision:) ?

resources/lemmas.md
daejunpark: to be honest, i'm not sure if the new arithmetic simplification lemmas are equivalent to the existing ones.  also, i'm not sure if the change -Int to +Int and the reverse +Int to -Int will not lead to the infinite loop.  but i hope that the infinite loop will be detected provided that we have no ambiguity of function rule application.

gnosis/test/verification.k
denis-bogdanas: Yeah' that's true. We actually need #isGround() - ensure that term doesn't have variables. Should we add a hook for it?

In the meantime, if all the rest is ok, can you merge please? This will help others get rid of non-determinism.

denis-bogdanas:
These normalization rules were enough to get rid of all non-determinisms. I agree that in theory they won't normalize every single expression. For example they won't fix A+B = B+A when A and B are symbolic. But they were good enough for now. And there were some normalization rules before present, they were even less universal than these, that's why I had do make changes. -Int to +Int normalization I think is beneficial. Without it there would be more particular cases.

gnosis/test/Makefile
```diff
@@ -6,7 +6,7 @@ TMPLS:=../module-tmpl.k spec-tmpl.k
 SPEC_GROUP:=gnosis-test
 SPEC_INI:=api-test.ini
 
-KPROVE_OPTS:=--smt-prelude $(abspath $(dir $(abspath $(lastword $(MAKEFILE_LIST))))/../evm.smt2)
+KPROVE_OPTS:=--smt-prelude $(abspath $(dir $(abspath $(lastword $(MAKEFILE_LIST))))/../evm.smt2) --deterministic-functions
```
daejunpark: why `--deterministic-functions` option only for this?   shouldn't we put it for all others?

resources/lemmas.md
denis-bogdanas: I took care of that. Infinite loop is not possible. These are the only 2 rules that rewrite a simpler expression into a more complex one:
```
    rule I -Int B => (0 -Int B) +Int I when #isConcrete(I) andBool notBool #isConcrete(B) andBool I =/=Int 0
    rule A -Int I => A +Int (0 -Int I) when notBool #isConcrete(A) andBool #isConcrete(I)
```
In first rule the side condition `I =/=Int 0` is essential to ensure rule doesn't apply indefinitely.
In 2nd rule a simplification will follow on the RHS. All other rules are crafted in such a way that they cannot match on the result of the 1st rule alone. And you are right we have the option to check determinism now that ensures nothing goes wrong.

gnosis/test/verification.k
```diff
@@ -117,14 +120,13 @@ module VERIFICATION
...
 
     // Arithmetic
-    rule I1 +Int (I2 +Int I3) => I2 +Int (I1 +Int I3) when #isConcrete(I1) andBool #isConcrete(I3)
-    rule I1 +Int (I2 +Int I3) => I3 +Int (I1 +Int I2) when #isConcrete(I1) andBool #isConcrete(I2)
-
-    rule (I1 +Int I2) +Int (I3 -Int I2) => I1 +Int I3
-    rule (I1 +Int I2) -Int (I1 +Int I3) => I2 -Int I3
+    rule (A +Int I2) +Int (I3 -Int A) => I2 +Int I3
+    rule (A +Int I2) -Int (A +Int I3) => I2 -Int I3
+    rule (A +Int I2) -Int (I3 +Int A) => I2 -Int I3
```
daejunpark: I'm asking why the removed four rules are equivalent to the new three rules.

gnosis/test/verification.k
daejunpark: That's not what i'm asking.  I started a new thread below.

gnosis/verification.k
daejunpark: well, we're talking about probability. you're assuming that the probability of getting a hash value that has ~220 leading zeros is negligible. someone agrees on that. but i'm saying that it is better to make the probability even smaller.  i mean, i think `I` that appears in the execution of gnosis should be less than 10 or 20, since they are all storage index for the global variables, right?  btw, it's not urgent, and we can do this later.

gnosis/test/verification.k
daejunpark: i lost.  you mean that both A and B should NOT have a variable inside?  i don't think it should be the case.  any storage access with a function parameter value as an index will lead to the hash2 term with variables.

gnosis-imap/gnosis-imap-spec.ini
```diff
@@ -0,0 +1,194 @@
...
+localMem: .Map => _
+pc: 0 => _
+gas: #infinity => _
+memoryUsed: 0 => _
+callDepth: CD
+log: _ => _
+refund: _ => _
+coinbase: _ => _
+; account 1 has to be active, otherwise there will be branching on <k>: #accountNonexistent(1)
+activeAccounts: SetItem(MASTER_COPY_ID) SetItem(PROXY_ID) SetItem(ORIGIN_ID) SetItem(1) _:Set
```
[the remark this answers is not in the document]
yzhang90: In fact, I can fix it in my PR.

resources/lemmas.md
daejunpark: hmm. @yzhang90 is right.  if the first rule is applied first, then we have `(A + I2) + ((0 - A) + I3)`, which i'm not sure how to be evaluated to `I2 + I3`.

anyway, it seems that we still have the ambiguity, and i think this type of ambiguity cannot be detected by the --deterministic-function feature, since the applicable terms are different. 

@denis-bogdanas just curious, what were the existing ambiguous rules you found initially?

daejunpark: @denis-bogdanas my guess is that the ambiguity came from the `infinity` rules initially, which we plan to drop soon, since the rules are not sound.  if that is the case, we can first drop the infinity rules and could minimize the changes to resolve the remaining ambiguity.  what do you think?

resources/lemmas.md
```diff
@@ -184,12 +183,19 @@ The rules are applied only when the side-conditions are met.
...
 
 ```k
-    rule (I1 +Int I2) +Int I3 => I1 +Int (I2 +Int I3) when #isConcrete(I2) andBool #isConcrete(I3)
-    rule (I1 +Int I2) -Int I3 => I1 +Int (I2 -Int I3) when #isConcrete(I2) andBool #isConcrete(I3)
-    rule (I1 -Int I2) +Int I3 => I1 -Int (I2 -Int I3) when #isConcrete(I2) andBool #isConcrete(I3)
-    rule (I1 -Int I2) -Int I3 => I1 -Int (I2 +Int I3) when #isConcrete(I2) andBool #isConcrete(I3)
+    //orienting symbolic term to be first, converting -Int to +Int
+    rule I +Int B => B          +Int I when #isConcrete(I) andBool notBool #isConcrete(B)
+    rule I -Int B => (0 -Int B) +Int I when #isConcrete(I) andBool notBool #isConcrete(B) andBool I =/=Int 0
+    rule A -Int I => A +Int (0 -Int I) when notBool #isConcrete(A) andBool #isConcrete(I)
```
daejunpark: it turned out that the above two rules are two strong (applicable in too many contexts).  every time we have some special form of symbolic terms, say `foo(X) - Y`, that we want to have its own simplification rule, we need to have either add more side-condition to the above rules to prevent them from being applied for the special terms, or we need to write a simplification rule for the new term that results from applying the above rule, i.e., `foo(X) + (-Y)`.

i'm wondering if the above two rules are really needed to fix the ambiguity, or you generalized all the way up, potentially being more than what is necessary, as you went through.

daejunpark: I strongly propose not to have the normalization from -Int to +Int, unless it is necessary.

Or, if it is really necessary, I think we may be able to drop many of these arithmetic simplification rules, since they came from simplifying the gas terms, but now we are using a new gas abstraction which we can control more.

What do you think guys? @denis-bogdanas @yzhang90

resources/lemmas.md
denis-bogdanas: Ambiguities didn't come only from #infinity. There were many other cases. I didn't keep record of all of them. I would just add new rules for the ambiguous terms that were discovered, until I got too many repetitive rules and I decided to do some low-level normalization instead, like converting -Int to +Int.

The most curious case was indeed related to #infinity. `#infinity -Int #infinity` could rewrite to either #infinity or 0 :D. But because that was part of gas availability logic, it probably didn't have any effect on overall determinism. I fixed this by adding side conditions like
```
    rule N -Int N => 0
      requires #getKLabelString(N) =/=String "#infinity_VERIFICATION"
```

Now to the question why there's no non-determinism in the rule pointed by @yzhang90 : I'm surprised myself. I guess that's because of shady ways kprove rewrite engine works.

If full evaluation is used, then kprove will first evaluate sub-terms then parent terms, and then the `0 -Int A` heating rule will apply and big simplification rule should not be able to apply. But if `substituteAndEvaluate` is used instead, which happens in most cases, only sub-terms affected by substitution are evaluated, and I guess because of this parent term gets evaluated first.

Even if non-determinism arises, it can be fixed by replacing the rule with
```
rule (A +Int I2) +Int ((0 -Int A) +Int I3) => I2 +Int I3
```

I'll activate `--deterministic-functions` for all gnosis projects, my guess is they will all be deterministic.

Anyway, there is one more normalization principle that can be applied and it would fix any possible issues due to rules above: group parentheses always to the right. That is:
```
rule (A+B)+C => A+(B+C) when A is symbolic.
```
This will ensure that symbolic terms accumulate at the left and concrete at the right, where they can add up. So far this was not needed.

I guess you can agree that mathematically lemmas are correct. If you ask me now to get rid of -Int to +Int normalization, I'd have to spend a couple more hours to add other lemmas for parentheses regrouping with +Int and -Int, and to rewrite arithmetic lemmas in verification.k into other forms that will inevitably occur.
So I propose let's not overthink this topic, leave the lemmas as is for now and extend/change only when the need arises. So that we can spend our time on actual gnosis specs. At least these lemmas have an advantage over the previous ones: they are deterministic (modulo kprove engine :D), and they make the specs pass both on our computers and on Jenkins.

gnosis/verification.k
denis-bogdanas: fixed

gnosis/test/verification.k
denis-bogdanas: Actually you are right. This side condition is not needed at all. Even if `#hash2(A,B)` is ground and should evaluate to a concrete number, we assume that number is very large, thus the last side condition ensures it cannot collide with I.

Deleted this side condition.

gnosis/test/Makefile
denis-bogdanas: fixed. For all gnosis specs. For others I'll open a separate PR just in case.

gnosis/verification.k
```diff
@@ -6,33 +6,74 @@ module VERIFICATION
...
+    rule storeRange(storeRange(M, MS0, MW0, #bufSeg(BUF, BS0, BW0)), MS1, MW1, #bufSeg(BUF, BS1, BW1))
+        => storeRange(M, MS0, MW0 +Int MW1, #bufSeg(BUF, BS0, MW0 +Int MW1))
+      requires #isBuf(BUF) andBool MW0 ==Int BW0 andBool MW1 ==Int BW1
+       andBool MS1 ==Int MS0 +Int MW0 andBool BS1 ==Int BS0 +Int BW0
+
+    // select/store key equality
+
+    rule hash2(A,B) ==K I => false
+      requires         notBool #isConcrete(hash2(A,B)) andBool #isConcrete(I)
+               andBool I <Int 1000000000 //random small-ish number bigger than any fixed storage location
```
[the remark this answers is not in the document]
denis-bogdanas: That was intentional, but fixed.

gnosis-imap/gnosis-imap-spec.ini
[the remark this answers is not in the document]
denis-bogdanas: 1. Discussed in the thread above.
2. Great, thanks!

resources/lemmas.md
denis-bogdanas: @daejunpark Ok, I removed -Int normalization for terms like `I -Int A`, but I kept it for `A -Int I`. Hope Jenkins is fine with that.

If I remove -Int normalziation entirely, but keep +Int reordering rule, then I get this mess:
```
[Error] Critical: [non-deterministic function definition]: more than one rule
can apply to the term:
_+Int_(
  Int(#"0"),,
  _-Int__INT(
    _*Int__INT(Int(#"65"),, NR_SIG_1903:Int),,
    Int(#"32")
  )
)

Candidate rules:
    rule I +Int B => B          +Int I when #isConcrete(I) andBool notBool
#isConcrete(B)
	Source: /mnt/d/verified-smart-contracts-fork/specs/lemmas.k
Location(187,10,187,90)

    rule I1 +Int (B -Int I3) => B +Int (I1 -Int I3) when #isConcrete(I1)
andBool notBool #isConcrete(B) andBool #isConcrete(I3)
	Source: /mnt/d/verified-smart-contracts-fork/specs/lemmas.k
Location(198,10,198,128)

Candidate results:
_-Int__INT(
  _*Int__INT(Int(#"65"),, NR_SIG_1903:Int),,
  Int(#"32")
)

_+Int_(
  _*Int__INT(Int(#"65"),, NR_SIG_1903:Int),,0
  Int(#"-32")
)
```

This will disappear if I also remove +Int reordering, but then we'll get a ton of issues elsewhere when 2 rules eventually rewrite something to a sum of terms but in a different order. So the current solution should be a good compromise.

If you still disagree please msg me on slack so that we can reach a consensus quickly.

denis-bogdanas: Jenkins results: now proxied-token fails. All gnosis projects pass. With previous version of lemmas all projects were passing. I disabled proxied-token to avoid spending time on this.

denis-bogdanas: @daejunpark @yzhang90 Actually a clean normalization solution would be to move symbolic terms of a sum to the RIGHT, not to the left:
```
rule A +Int I => I          +Int A when #isConcrete(I) andBool notBool #isConcrete(B)
```
That's because  both +B and -B can stay at the right-end of a sum, but only +B can stay at the left end. Damn it. It would have all advantages of my initial solution and no disadvantages. Should i do this change?

PR #94, opened 2018-12-05
TEST: Smtlib experiment

denis-bogdanas:
1. Reverted to older K-gnosis with original Z3 encoding algorithm.
2. KEVM: Marked with [smtlib] #symEcrec and `Int ":" WordStack`.


denis-bogdanas:
@daejunpark I was actually going to fix this PR only to see what happens. But of course there are more important things to do. Are we done with this experiment? Should I delete all the branches created for the sake of it?

PR #96, opened 2018-12-08
Gnosis test

daejunpark:


daejunpark:
@denis-bogdanas  Please review.

1. I'd like to merge this smaller PR only for the encodeTransactionData.
2. I added a new goal `jenkins` to Makefile, which runs each directory one-by-one (with cleaning up both K and KEVM at the end of each directory execution).  This seems work, avoiding the Jenkins failure due to the presence of multiple versions of K/KEVM.

For your other PRs, please cherry-pick the Makefile change, as the Jenkins setting is updated to use the `jenkins` goal.
https://github.com/runtimeverification/verified-smart-contracts/pull/96/commits/0a7a3e6d09c3bb80fd24b96de5b3ca92a626352b

daejunpark:
BTW, note that I dropped the K_OPTS export in Makefile, as the default value (12g) is too large for many machines including the Travis-CI server.  We can always call Makefile with setting up K_OPTS, like:
```
$ make K_OPTS=...
```
or
```
$ export K_OPTS=...
$ make
```

daejunpark:
subsumed by #93

daejunpark:
@denis-bogdanas 

The current Jenkins is set to 12GB. Do you want to increase or decrease?

> PIease set larger memory usage on Jenkins. 8 GB should be enough for what we have + some headroom. After I migrated Gnosis to IMap there are 2x less execution paths on larger specs.

denis-bogdanas:
Anything that makes tests pass. 8 GB should be enough.

PR #97, opened 2018-12-11
zeppelin and vyper with IMap

denisadiaconescu:
These changes were tested with the K and EVM commits specified in the .k.rev and .kevm.rev files.


erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -71,56 +74,66 @@ requires:
...
-    _:Map
+storage: M1:IMap => M2:IMap
+origStorage: M1
 +requires:
+    andBool select(M1, #hashedLocation({COMPILER}, {_BALANCES}, CALLER_ID)) ==Int BAL_FROM
     andBool CALLER_ID ==Int TO_ID
     andBool VALUE >Int BAL_FROM
+ensures:
+    ensures M1 ==IMap M2 
+        except (SetItem(#hashedLocation({COMPILER}, {_BALANCES}, CALLER_ID)) .Set)
```
daejunpark: why do you need this `except` here? the original spec says that the storage does NOT update at all.

erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -131,60 +144,122 @@ requires:
...
+       SetItem(#hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) .Set)
+
+[transferFrom-success-1-b]
++requires:
+    andBool #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) ==Int #hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)
+    andBool #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) =/=Int #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)
+ensures:
+    ensures select(M2, #hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)) ==Int BAL_FROM -Int VALUE
+    andBool select(M2, #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)) ==Int BAL_TO +Int VALUE
+    andBool select(M2, #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) ==Int ALLOW -Int VALUE
```
daejunpark: how could this possible where we have the hash collision?

daejunpark: oh, BAL_FROM is equal to ALLOW

daejunpark: i think it is better to explicitly say that BAL_FROM ==Int ALLOW in the ensures, to highlight it.

erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -131,60 +144,122 @@ requires:
...
+       SetItem(#hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) .Set)
+
+[transferFrom-success-1-c]
++requires:
+    andBool #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) =/=Int #hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)
+    andBool #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) ==Int #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)
+ensures:
+    ensures select(M2, #hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)) ==Int BAL_FROM -Int VALUE
+    andBool select(M2, #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)) ==Int BAL_TO -Int VALUE
+    andBool select(M2, #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) ==Int ALLOW -Int VALUE
```
daejunpark: same here, BAL_TO ==Int ALLOW in ensures.

daejunpark: not much sense to name it as success case.  you need to put a nice comment what this spec is trying to say.  without careful look, no one will even recognize BAL_TO -Int VALUE.

daejunpark: not only for this one, but for all "success-1-*" cases.

erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -131,60 +144,122 @@ requires:
...
 +requires:
+    andBool select(M1, #hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)) ==Int BAL_FROM
+    andBool select(M1, #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) ==Int ALLOW
     andBool FROM_ID ==Int TO_ID
     andBool VALUE <=Int BAL_FROM
     andBool VALUE <=Int ALLOW
+ensures:
+    ensures select(M2, #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) ==Int ALLOW -Int VALUE
+    andBool M1 ==IMap M2 except
+      (SetItem(#hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) .Set)
```
daejunpark: no hash collision here? why?

erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -131,60 +144,122 @@ requires:
...
-    #hashedLocation({COMPILER}, {_BALANCES},   FROM_ID)           |-> (BAL_FROM => _)  // BAL_FROM
-    #hashedLocation({COMPILER}, {_BALANCES},   TO_ID)             |-> (BAL_TO   => _)  // BAL_TO
-    #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) |-> ALLOW
-    _:Map
+storage: M1:IMap => M2:IMap
+origStorage: M1
++requires:
+    andBool select(M1, #hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)) ==Int BAL_FROM
+    andBool select(M1, #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)) ==Int BAL_TO
+    andBool select(M1, #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) ==Int ALLOW
```
daejunpark: nitpick: follow the vertical alignment.  see the original spec for the storage.

erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -131,60 +144,122 @@ requires:
...
+         SetItem(#hashedLocation({COMPILER}, {_BALANCES}, TO_ID)) .Set)
+
+[transferFrom-failure-1-b]
++requires:
+    andBool FROM_ID =/=Int TO_ID
+    andBool VALUE <=Int BAL_FROM
+    andBool BAL_TO +Int VALUE <Int (2 ^Int 256)
+    andBool VALUE <=Int ALLOW
+    andBool #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) ==Int #hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)
+    andBool #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) ==Int #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)
```
daejunpark: how could the allowance location be the same with two different locations?

erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -131,60 +144,122 @@ requires:
...
+[transferFrom-failure-1-a]
 +requires:
     andBool FROM_ID =/=Int TO_ID
     andBool ( VALUE >Int BAL_FROM
      orBool   BAL_TO +Int VALUE >=Int (2 ^Int 256)
      orBool   VALUE >Int ALLOW )
+ensures:
+    ensures M2 ==IMap M1 except 
+        (SetItem(#hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)) 
+         SetItem(#hashedLocation({COMPILER}, {_BALANCES}, TO_ID)) .Set)
```
daejunpark: no hash collision stuff here?

erc20/vyper/vyper-erc20-spec.ini
daejunpark: i cannot understand these two cases: failure-1-[ab].
you are better to explain this nontrivial case split.

erc20/zeppelin/zeppelin-erc20-spec.ini
```diff
@@ -145,54 +159,112 @@ requires:
...
+    andBool select(M2, #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) ==Int ALLOW
+   andBool M1 ==IMap M2 except
+     (SetItem(#hashedLocation({COMPILER}, {_BALANCES}, FROM_ID))
+      SetItem(#hashedLocation({COMPILER}, {_BALANCES}, TO_ID)) 
+      SetItem(#hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) .Set)
 
+[transferFrom-success-1-d]
++requires:
+    andBool #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) ==Int #hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)
+    andBool #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) ==Int #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)
```
daejunpark: i don't see how this is possible. we have a lemma saying that two hash2 terms are different if the arguments are different, and here we know that FROM_ID =/= TO_ID.

resources/lemmas.md
```diff
@@ -173,6 +173,9 @@ It reduces the reasoning efforts of the underlying theorem prover, factoring out
     rule byteStack2IntList ( WS ) => byteStack2IntList ( WS , #sizeWordStack(WS) /Int 32 ) requires #sizeWordStack(WS) %Int 32 ==Int 0
     rule byteStack2IntList ( WS , N ) => #asWord ( WS [ 0 .. 32 ] ) byteStack2IntList ( #drop(32, WS) , N -Int 1 ) requires N >Int 1
     rule byteStack2IntList ( WS , 1 ) => #asWord ( WS [ 0 .. 32 ] ) .IntList
+
+    rule hash2(K1,V1) =/=Int hash2(K2,V2) => K1 =/=Int K2 orBool V1 =/=Int V2
+    rule hash2(K1,V1) ==Int hash2(K2,V2) => K1 ==Int K2 andBool V1 ==Int V2
```
daejunpark: this lemma does NOT make sense if you want to explicitly reason about the hash collision.

you have to be consistent, choosing either
- explicit reasoning about the hash collision without this lemma
- assuming that no hash collision with or without this lemma

daejunpark:
@denisadiaconescu overall looks ok, but i'm not happy with the hash collision reasoning.  i'm ok with explicit hash collision reasoning, but you should be consistent.  both explicit reasoning and having the hash2 lemma does NOT make sense.  see the detailed comments for each lines, and please address them.

denisadiaconescu:
Jenkins: test this please

erc20/vyper/vyper-erc20-spec.ini
denisadiaconescu: I fixed this.

denisadiaconescu:
@daejunpark Thanks for the review. I had the impression that I deleted the lemma for `hash2(K1,V1) ==Int hash2(K2,V2)`, sorry for this. Now I deleted the lemma, but I also dropped the partial hash collision analysis. 

The only major change now is in `transferFrom-success-1` (for both `vyper` and `zeppelin`) in which we need to add two extra conditions (that some hash values are different) in order to prove that case.

denisadiaconescu:
@daejunpark Thanks for the review. Sorry about that lemma, I had the impression that I deleted it and I was not trying to use it. Now I deleted the lemma, but also dropped the (partial) hash collision analysis. The main changes now are in `transferFrom-success-1` (for both `vyper` and `zeppelin`) in which we added two extra conditions (saying explicitly that some hash values are not equal) in order to prove that case. 

I cherry-picked the commit you pointed out, but Jenkins still fails with the error `No rule to make target 'jenkins'`.

Makefile
```diff
@@ -97,8 +97,8 @@ zeppelin_erc20_files:=totalSupply-spec.k \
              transferFrom-success-2-spec.k \
              transferFrom-failure-1-a-spec.k \
              transferFrom-failure-1-b-spec.k \
-             transferFrom-failure-2-spec.k
-
+             transferFrom-failure-2-spec.k 
+
```
daejunpark: we shouldn't have this non-necessary white-space changes. please fix it.

erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -71,56 +79,71 @@ requires:
...
-    #hashedLocation({COMPILER}, {_BALANCES}, CALLER_ID) |-> (BAL_FROM => BAL_FROM -Int VALUE)
-    #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)     |-> (BAL_TO   => BAL_TO   +Int VALUE)
-    _:Map
+storage: 
+    M1:IMap => M2:IMap
+origStorage: 
+    M1
 +requires:
+    andBool select(M1, #hashedLocation({COMPILER}, {_BALANCES}, CALLER_ID)) ==Int BAL_FROM
+    andBool select(M1, #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)) ==Int BAL_TO
```
daejunpark: by alignment, i mean, align `==Int`, by adding some white spaces between `TO_ID))` and `==Int`.

erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -71,56 +79,71 @@ requires:
...
     andBool CALLER_ID =/=Int TO_ID
     andBool VALUE <=Int BAL_FROM
-    andBool BAL_TO +Int VALUE <Int (2 ^Int 256)
-
+    andBool BAL_TO +Int VALUE <Int (2 ^Int 256)  
+ensures:
+    ensures select(M2, #hashedLocation({COMPILER}, {_BALANCES}, CALLER_ID)) ==Int BAL_FROM -Int VALUE
+    andBool select(M2, #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)) ==Int BAL_TO +Int VALUE
```
daejunpark: alignment here as well

erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -71,56 +79,71 @@ requires:
...
-    #hashedLocation({COMPILER}, {_BALANCES}, CALLER_ID) |-> (BAL_FROM => _)
-    #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)     |->  BAL_TO
-    _:Map
+storage: 
+    M1:IMap => M2:IMap
+origStorage: 
+    M1
 +requires:
+    andBool select(M1, #hashedLocation({COMPILER}, {_BALANCES}, CALLER_ID)) ==Int BAL_FROM
+    andBool select(M1, #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)) ==Int BAL_TO
```
daejunpark: align

erc20/vyper/vyper-erc20-spec.ini
```diff
@@ -131,60 +154,85 @@ requires:
...
 +requires:
+    andBool select(M1, #hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)) ==Int BAL_FROM
+    andBool select(M1, #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)) ==Int BAL_TO
+    andBool select(M1, #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID)) ==Int ALLOW
     andBool FROM_ID =/=Int TO_ID
     andBool VALUE <=Int BAL_FROM
     andBool BAL_TO +Int VALUE <Int (2 ^Int 256)
     andBool VALUE <=Int ALLOW
+    andBool #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) =/=Int #hashedLocation({COMPILER}, {_BALANCES}, FROM_ID)
+    andBool #hashedLocation({COMPILER}, {_ALLOWANCES}, FROM_ID CALLER_ID) =/=Int #hashedLocation({COMPILER}, {_BALANCES}, TO_ID)
```
daejunpark: this doesn't look good.  you have to add this kind of requires for all specs, without using the lemma at all,  or no this kind of requires with using the lemma only.

erc20/vyper/vyper-erc20-spec.ini
daejunpark: no you didn't.  what did you fix?

denisadiaconescu:
Created new PR to `master` #99

PR #98, opened 2018-12-12
Gas imap

yzhang90:


gnosis-imap/gnosis-imap-spec.ini
```diff
@@ -6,27 +6,17 @@
...
-callStack: _
-wordStack: .WordStack => _
-localMem: .Map => _
-pc: 0 => _
-gas: #infinity => _
-memoryUsed: 0 => _
 callDepth: CD
-log: _ => _
-refund: _ => _
-coinbase: _ => _
```
daejunpark: why did you remove this? you can just overwrite by specifying them in your spec.

daejunpark: maybe you think that this helps the readability and less error-prone?

gnosis-imap/gnosis-imap-spec.ini
```diff
@@ -35,76 +25,49 @@ ensures:
...
 this: PROXY_ID
 msg_sender: MSG_SENDER
-callData: #abiCallData("signatureSplit", #toBytes(SIGNATURES, 65 *Int NR_SIG), #uint256(POS))
+callData: _
 callValue: 0
+wordStack: POS : SIGS_LOC : RETURN_LOC : WS =>
+           RETURN_LOC : #asWord(#bufSeg(SIGS_BUF, 65 *Int POS +Int 32, 32)) : #asWord(#bufSeg(SIGS_BUF, 65 *Int POS, 32)) : #bufElm(SIGS_BUF, 65 *Int POS +Int 64) : WS
+pc: 22163 => 22209
+gas: #gas(NONMEMGAS, MEMGAS) => #gas(NONMEMGAS +Int 111, MEMGAS +Int (Cmem(BYZANTIUM, #memoryUsageUpdate(MU, SIGS_LOC +Int (65 *Int POS +Int 65), 32)) -Int Cmem(BYZANTIUM, MU)))
+memoryUsed: MU => #memoryUsageUpdate(MU, SIGS_LOC +Int (65 *Int POS +Int 65), 32)
```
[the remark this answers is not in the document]
yzhang90: You mean add a substitution in the require clause?

gnosis-imap/gnosis-imap-spec.ini
```diff
@@ -130,6 +94,14 @@ callData: #abiCallData("encodeTransactionData", (
             #address(REFUND_RECEIVER),
             #uint256(NONCE) ))
 callValue: 0
+wordStack: .WordStack => _
+localMem: .IMap => _
+pc: 0 => _
+gas: #gas(0,0) => _
+memoryUsed: 0 => _
```
daejunpark: this will be the case even if this function is called inside this contract?
i mean, when execTransaction calls encodeTransactionData, it uses the external call mechanism by spawning a new VM, or just jump to the function locally?
i guess `public` and `external` functions are different...?

yzhang90: https://ethereum.stackexchange.com/questions/19380/external-vs-public-best-practices

gnosis/verification.k
```diff
@@ -6,21 +6,35 @@ module VERIFICATION
...
+    rule #memoryUsageUpdate(#memoryUsageUpdate(MU, START0, WIDTH0), START1, WIDTH1) => #memoryUsageUpdate(MU, START0, WIDTH0)
+      requires START1 +Int WIDTH1 <Int START0 +Int WIDTH0
+
+    rule #memoryUsageUpdate(#memoryUsageUpdate(MU, START0, WIDTH0), START1, WIDTH1) => #memoryUsageUpdate(MU, START1, WIDTH1)
+      requires START0 +Int WIDTH0 <=Int START1 +Int WIDTH1
+
+    rule 0 <=Int #memoryUsageUpdate(MU, START, WIDTH) => true
+
+    rule pow256 <=Int #memoryUsageUpdate(MU, START, WIDTH) => false  requires START +Int WIDTH <Int 3705346855594118253554271520278013051304639509300498049262642688253220148477921
+    rule #memoryUsageUpdate(MU, START, WIDTH) <Int  pow256 => true   requires START +Int WIDTH <Int 3705346855594118253554271520278013051304639509300498049262642688253220148477921
```
daejunpark: could you put the comment explaining what this big number means?

gnosis/verification.k
```diff
@@ -6,21 +6,35 @@ module VERIFICATION
...
 
+  // ########################
+  // Memory Usage
+  // ########################
+
+    rule #memoryUsageUpdate(#memoryUsageUpdate(MU, START0, WIDTH0), START1, WIDTH1) => #memoryUsageUpdate(MU, START0, WIDTH0)
+      requires START1 +Int WIDTH1 <Int START0 +Int WIDTH0
+
+    rule #memoryUsageUpdate(#memoryUsageUpdate(MU, START0, WIDTH0), START1, WIDTH1) => #memoryUsageUpdate(MU, START1, WIDTH1)
+      requires START0 +Int WIDTH0 <=Int START1 +Int WIDTH1
```
daejunpark: for both rules, i think you should reason about the case when WIDTH0 or WIDTH1 is zero.

my suggestion (not sure if works though):
- adding WIDTH0 >Int 0 andBool WIDTH1 >Int 0
- adding another rule:
```
rule #memoryUsageUpdate(MU, START, WIDTH) => MU  when WIDTH ==Int 0
```
or change the semantic rule directly:
https://github.com/kframework/evm-semantics/pull/281/files#diff-1e76ae9645a9718bd96a177a24eb3865R2000

gnosis/verification.k
```diff
@@ -6,21 +6,35 @@ module VERIFICATION
...
- // -------------------------------------
-    rule #infinity +Int _ => #infinity
-    rule #infinity -Int _ => #infinity
-    rule A <=Int #infinity => true
-      requires #getKLabelString(A) =/=String "#infinity_VERIFICATION"
-    rule #infinity <Int A => false
-      requires #getKLabelString(A) =/=String "#infinity_VERIFICATION"
+    syntax Int ::= #gas ( Int , Int ) [function]
+ // --------------------------------------------
+    rule #gas(NONMEM, MEM) -Int #symCmem(MEM') => #gas(NONMEM, MEM +Int MEM')
```
daejunpark: why `MEM +Int MEM'` not `MEM -Int MEM'` ?

daejunpark: ah, #gas(...) is the accumulated amount of deducted gas?

gnosis/verification.k
```diff
@@ -6,21 +6,35 @@ module VERIFICATION
...
+    rule #gas(NONMEM, MEM) +Int NONMEM' => #gas(NONMEM -Int NONMEM', MEM)  requires #getKLabelString(NONMEM') =/=String "#symCmem"
+    rule #gas(NONMEM, MEM) -Int NONMEM' => #gas(NONMEM +Int NONMEM', MEM)  requires #getKLabelString(NONMEM') =/=String "#symCmem"
+
+    rule G <=Int #gas(_,_)  => true  requires #getKLabelString(G) =/=String "#gas"
+    rule #gas(_, _) <Int G  => false requires #getKLabelString(G) =/=String "#gas"
```
daejunpark: i don't understand these two rules.

gnosis/verification.k
yzhang90: We are accumulating all the gas during the execution, right?  MEM' ==K Cmem(New) -Cmem(OLD) is the gas cost. So we accumulate the cost in the mem part of #gas .

gnosis/verification.k
```diff
@@ -67,14 +81,17 @@ module VERIFICATION
 
     rule WS ++ .WordStack => WS
 
+    rule sizeWordStackAux(WS, N) <Int SIZE => sizeWordStackAux(WS, 0) +Int N <Int SIZE  requires N =/=Int 0
+    rule SIZELIMIT <Int sizeWordStackAux(WS, N) +Int DELTA  => SIZELIMIT <Int (sizeWordStackAux(WS, 0) +Int N) +Int DELTA  requires N =/=Int 0
+    rule SIZELIMIT <Int sizeWordStackAux(WS, N) => SIZELIMIT <Int sizeWordStackAux(WS, 0) +Int N  requires N =/=Int 0
```
daejunpark: could you align these two rules so that we can see the difference quickly?

gnosis/verification.k
yzhang90: Used to solve the side condition for these two rules.
```   rule <k> G:Int ~> #deductGas => #end EVMC_OUT_OF_GAS ... </k> <gas> GAVAIL                  </gas> requires GAVAIL <Int G
    rule <k> G:Int ~> #deductGas => .                    ... </k> <gas> GAVAIL => GAVAIL -Int G </gas> <previousGas> _ => GAVAIL </previousGas> requires GAVAIL >=Int G```
GAVAIL will be #gas(_,_) during the execution.

File: gnosis/verification.k [hunk 9]
```diff
@@ -67,14 +81,17 @@ module VERIFICATION
...
+
+    rule #sizeWordStack(selectRange(_, _, WIDTH), SIZE) => SIZE +Int WIDTH
+
 
   // ########################
   // Range
   // ########################
 
-    //todo should not be needed since #bufSeg(_,_,1) is converted to #bufElm
-    rule #asWord(#bufSeg(_, _, WIDTH)) <Int 256 => true requires WIDTH ==Int 1
```
daejunpark: are you sure this is not needed? or any particular reason not to have this?

gnosis/verification.k
```diff
@@ -99,31 +116,69 @@ module VERIFICATION
     rule 0 <=Int #bufElm(_, _)  => true
     rule #bufElm(_, _) <Int 256 => true
 
+    rule 0 <=Int X xorInt 115792089237316195423570985008687907853269984665640564039457584007913129639935 => true      requires 0<=Int X andBool X <Int pow256
+
+    rule X xorInt 115792089237316195423570985008687907853269984665640564039457584007913129639935 <Int pow256 => true  requires 0<=Int X andBool X <Int pow256
```
daejunpark: please put a comment for the big magic number.

gnosis/verification.k
yzhang90:
``` 
    rule #bufSeg(BUF, START, 1) => #bufElm(BUF, START) : .WordStack
    rule #asWord(#bufElm(BUF, INDEX) : .WordStack) => #bufElm(BUF, INDEX)
    rule #bufElm(_, _) <Int 256 => true
```

gnosis/verification.k
```diff
@@ -99,31 +116,69 @@ module VERIFICATION
...
+    rule #buf(32, DATA) => #padToWidth(32, #asByteStack(DATA))  [concrete]
 
     rule #asWord(WS0 ++ #buf(MASKWIDTH0, 0)) |Int #asWord(#buf(MASKWIDHT1, 0) ++ WS1)
       => #asWord(WS0 ++ WS1)
       requires #sizeWordStack(WS0) +Int MASKWIDTH0 ==Int 32
        andBool #sizeWordStack(WS0) ==Int MASKWIDHT1
        andBool MASKWIDTH0 ==Int #sizeWordStack(WS1)
 
-    // Arithmetic
+    rule #asWord(WS) &Int 115792089237316195423570985008687907853269984665640564039457584007913129639935 => #asWord(WS)
```
daejunpark: again, please comment for the big number

gnosis/verification.k
yzhang90:
```
   rule #bufSeg(BUF, START, 1) => #bufElm(BUF, START) : .WordStack
    rule #asWord(#bufElm(BUF, INDEX) : .WordStack) => #bufElm(BUF, INDEX)
    rule #bufElm(_, _) <Int 256 => true
```

gnosis/verification.k
```diff
@@ -99,31 +116,69 @@ module VERIFICATION
...
+
+    // 2^256 - 2^160 = 0xff..ff00..00 (96 1's followed by 160 0's)
+    rule 115792089237316195423570985007226406215939081747436879206741300988257197096960 &Int ADDR => 0
+      requires #rangeAddress(ADDR)
+
+    rule 1 |Int chop(A xorInt maxUInt256) => 1
```
daejunpark: why?

yzhang90: I didn't add these two rules. It was there in the verification.k long ago and I just moved it around

gnosis/verification.k
```diff
@@ -99,31 +116,69 @@ module VERIFICATION
...
+  // Arithmetic
+  // ########################
+
     rule (A +Int I2) +Int (I3 -Int A) => I2 +Int I3
     rule (A +Int I2) -Int (A +Int I3) => I2 -Int I3
     rule (A +Int I2) -Int (I3 +Int A) => I2 -Int I3
 
+    rule (I1 +Int I2) -Int (I3 +Int I1) => I2 -Int I3
+
+    rule A -Int (A -Int I1) => I2
```
daejunpark: typo: I2 => I1

gnosis/verification.k
```diff
@@ -99,31 +116,69 @@ module VERIFICATION
...
+
+    // Simplify Cmem(_,_) - Cmem(_, _)
+    rule A +Int (I1 -Int A) => I1
+    rule (I1 +Int (A -Int I0)) +Int (I2 -Int A) => I1 +Int (I2 -Int I0)
```
daejunpark: don't we need to put more side condition so that these rules are applied in a controlled way?
e.g., isConcrete(I)  and/or  getKLabel(A) == "Cmem" ...?

yzhang90: Not sure. Eliminating the common expression is always preferred, isn't it?

gnosis-imap/gnosis-imap-spec.ini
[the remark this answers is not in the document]
daejunpark: yes, but not necessarily now, (since it may not work for unknown reason)

gnosis-imap/gnosis-imap-spec.ini
daejunpark: so, this spec cannot be used when proving execTransaction, right?

gnosis/verification.k
daejunpark: i mean, we need to isolate this rule application, so that it is applied only for the semantic rules you mentioned.  otherwise, these rules are not sound.

you may need to change the semantics as well, in order to isolate these.

gnosis/verification.k
daejunpark: could you comment it out and try?  this seems definitely unsound..

gnosis/verification.k
daejunpark: maybe you're right in this case.  i'm talking about some general principle.

gnosis/verification.k
daejunpark: ok.

gnosis/verification.k
daejunpark: is this #gas(...) stored in <gas> cell, which denotes the remaining gas, right?  then, #gas(...) should take three arguments, say #gas(ORIGIN, NONMEM, MEM), which denotes ORIGIN - NONMEM - MEM ?

gnosis/verification.k
daejunpark: following the above question, if we change #gas takes three arguments, #gas(ORIGIN, NONMEM, MEM), then, we can say like:
```
G <= Int #gas(#infinity, _, _) => true
```

gnosis-imap/gnosis-imap-spec.ini
yzhang90: You're correct

gnosis/verification.k
yzhang90: you're correct. But what value do we put for ORIGIN? a fresh variable?

daejunpark: like `#gas(#infinity, 0, 0)` ?

gnosis-imap/gnosis-imap-spec.ini
daejunpark: hmm, then this is not part of our verification scope..   we are interested in this function only in the context of execTransaction.

gnosis/verification.k
daejunpark: what i mean by `#gas(#infinity, 0, 0)`, is the LHS of the spec.

for the ORIGIN, i think we need to introduce `#infinity` again, but without any lemma directly over it.

gnosis/verification.k
yzhang90: will #infinity ever go out side #gas(#infinity, _, _) expression? why we need getKLabel(G) =/= #infinity?

daejunpark: i don't know, i'm just giving an example.

my biggest concern for this is when we call an external function (where we pass the currently available gas) and return from it (where we receive the remaining gas).  then, `G` could be`#gas(...)` or some complex term that includes #gas(...) inside.  then, what should be the correct lemma for them?

gnosis/verification.k
yzhang90: done. see https://github.com/kframework/evm-semantics/commit/37a90febd2e727e4fded639e6cada074050cc5e4

yzhang90:
jenkins: test this please.

gnosis/verification.k
yzhang90: No longer needed. see https://github.com/kframework/evm-semantics/commit/da13d241f6521766a240227b00d1d02b7bf5a6ca

gnosis/verification.k
yzhang90: done

resources/lemmas.md
```diff
@@ -328,13 +328,6 @@ These lemmas abstract some properties about `#sizeWordStack`:
 
 ```k
     rule 0 <=Int #sizeWordStack ( _ , _ ) => true [smt-lemma]
-    /*rule #sizeWordStack ( WS , N:Int )
-      => #sizeWordStack ( WS , 0 ) +Int N
-      requires N =/=K 0*/
-    rule #sizeWordStack ( WS , N:Int )
-      => N +Int #sizeWordStack ( WS , 0 )
-      requires N =/=K 0
-      [lemma]
```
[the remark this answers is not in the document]
yzhang90: done

gnosis/verification.k
yzhang90: fixed

gnosis/verification.k
yzhang90: deleted

gnosis/verification.k
yzhang90: done

gnosis/verification.k
yzhang90: done

gnosis-imap/gnosis-imap-spec.ini
```diff
@@ -169,6 +140,14 @@ proxy_storage:
                                  #address(REFUND_RECEIVER),
                                  #uint256(NONCE) ))
 
+[encodeTransactionData-1]
++requires:
+    andBool DATA_LEN ==Int 0
+
+[encodeTransactionData-2]
++requires:
+    andBool DATA_LEN =/=Int 0
```
daejunpark: `DATA_LEN =/=Int 0`  vs  `DATA_LEN ==Int 0`?

daejunpark:
@yzhang90 I'll merge this.  But note that I have a few more new comments. Please address them, and make a separate PR if needed.

gnosis-imap/gnosis-imap-spec.ini
```diff
@@ -35,86 +25,59 @@ ensures:
...
+memoryUsed: MU => FINAL_MEM_USAGE
+log: _
+refund: _
+coinbase: _
 proxy_storage:
-    _:Map
+    _:IMap
 +requires:
-    // Range
-    andBool #rangeAddress(PROXY_ID)
```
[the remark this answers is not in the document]
yzhang90: The requirement of #PROXY_ID is specified in the spec-tmpl.k

gnosis/abstract-semantics.k
```diff
@@ -1,10 +1,49 @@
...
+   rule <k> Ccallgas(SCHED, ISEMPTY:Bool, GCAP, GAVAIL, VALUE)
+          => #gas(Cgascap(SCHED, GCAP, GAVAIL, Cextra(SCHED, VALUE, ISEMPTY)), 0, 0) ... </k>
+     requires VALUE ==Int 0
+      andBool #getKLabelString(GCAP) ==String "#gas"
+      andBool GCAP ==Int GAVAIL
+     [trusted]
+
+   rule <k> Ccallgas(SCHED, ISEMPTY:Bool, GCAP, GAVAIL, VALUE)
+          => #gas(Cgascap(SCHED, GCAP, GAVAIL, Cextra(SCHED, VALUE, ISEMPTY)) +Int Gcallstipend < SCHED >, 0, 0) ... </k>
+     requires VALUE =/=Int 0
```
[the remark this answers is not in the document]
yzhang90: No specific reason. shall I change it to VALUE >Int 0?

PR #101, opened 2018-12-19
small fix: change =/=Int to <Int

yzhang90:


gnosis/abstract-semantics.k
```diff
@@ -36,7 +36,7 @@ module ABSTRACT-SEMANTICS
 
    rule <k> Ccallgas(SCHED, ISEMPTY:Bool, GCAP, GAVAIL, VALUE)
           => #gas(Cgascap(SCHED, GCAP, GAVAIL, Cextra(SCHED, VALUE, ISEMPTY)) +Int Gcallstipend < SCHED >, 0, 0) ... </k>
-     requires VALUE =/=Int 0
+     requires 0 <Int VALUE
```
daejunpark: this is better to have since it ensures that we have a sufficient pre-condition.

gnosis-imap/gnosis-imap-spec.ini
```diff
@@ -146,7 +146,7 @@ proxy_storage:
 
 [encodeTransactionData-2]
 +requires:
-    andBool DATA_LEN =/=Int 0
+    andBool 0 <Int DATA_LEN
```
daejunpark: logically, this makes no difference, but good to have for readability.

PR #103, opened 2018-12-20
Gnosis swapOwner-success 

anvacaru:
Current status on swapOwner-success case when all owners (prevOwner, oldOwner and newOwner) are different for gnosis contract


daejunpark:
#105

PR #107, opened 2018-12-21
Add owner with threshold

ana-pantilie:
I did the following:
  1. I added a (commented) sketch of the whole function
  2. I managed to prove the first success case with postconditions regarding `ownercount` and `threshold`. The storage contains the functions `chop()` and `&Int()`, I need to see where these come from and find a way to solve this in order to prove the spec with the rest of the postconditions.

### DIFF
--- a/gnosis-imap/gnosis-imap-spec.ini
+++ b/gnosis-imap/gnosis-imap-spec.ini
@@ -148,6 +148,138 @@ proxy_storage:
 +requires:
     andBool 0 <Int DATA_LEN
 
+;[addOwnerWithThreshold]
+;output: _ => _
+;callData: #abiCallData("addOwnerWithThreshold", (#address(OWNER), #uint256(NEW_THRESHOLD)))
+;callValue: 0
+;this: PROXY_ID
+;proxy_storage:
+;    M1:IMap => M2:IMap
+;msg_sender: MSG_SENDER
+;+requires:
+;   andBool #rangeAddress(MSG_SENDER)
+;   andBool #rangeAddress(PROXY_ID)
+;   andBool #rangeAddress(OWNER)
+;   andBool #rangeUInt(256, NEW_THRESHOLD)
+;   andBool select(M1, #hashedLocation({COMPILER}, {THRESHOLD},  .IntList)) ==Int THRESHOLD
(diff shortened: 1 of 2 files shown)


daejunpark:
I suspect that the `chop` and `&Int` issue is caused because you didn't provide the range condition for the `select(M1, K)` terms where `K` is an owner address.

Why don't you follow my specification example, specifying the storage entries explicitly?

gnosis-imap/gnosis-imap-spec.ini
```diff
@@ -148,6 +148,138 @@ proxy_storage:
...
+coinbase: _ => _
+output: _ => _
+this: #PROXY_ID
+msg_sender: #PROXY_ID
+callData: #abiCallData("addOwnerWithThreshold", (
+            #address(OWNER), 
+            #uint256(NEW_THRESHOLD) ))
+callValue: 0
+proxy_storage:
+    M1:IMap => M2:IMap
```
daejunpark: @ana-pantilie I'm not positive for this approach. Why don't you try my approach that explicitly specifies the `store` terms in the LHS?  Please let me know if you have a specific reason to use this approach.

ana-pantilie:
Thank you for the review, Daejun. As suggested, I wrote the structure of the store explicitly in proxy_storage. I'm not sure why this is needed, but I'll try to understand this after I finish proving the function.

Indeed, I forgot to add the range condition for `OLD_OWNER`. Now the storage looks fine, and the spec is proved with postconditions regarding `OWNER` and `OLD_OWNER` as well. The spec fails with the `==IMap except` condition. I've been trying to figure out why, but I'm stuck. The storage looks as it should, and from the definition of `==IMap except`, the condition should rewrite to true, unless I'm missing something.

gnosis-imap/gnosis-imap-spec.ini
```diff
@@ -148,6 +148,134 @@ proxy_storage:
...
+;ensures:
+;
+;[addOwnerWithThreshold-failure-2]
+;statusCode: _ => EVMC_INVALID_INSTRUCTION
+;log: _
+;+requires:
+;    andBool (OWNERCOUNT +Int 1) >=Int (2 ^Int 256)
+;    andBool OWNER =/=Int 0
+;    andBool OWNER =/=Int 1
+;ensures:
```
daejunpark: let's drop this, keeping in your local copy just in case.

PR #110, opened 2018-12-24
removeOwner failure cases

denisadiaconescu:


gnosis-imap/gnosis-imap-spec.ini
```diff
@@ -172,19 +171,27 @@ proxy_storage:
                                      #hashedLocation({COMPILER}, {THRESHOLD}, .IntList), THRESHOLD),
                                      #hashedLocation({COMPILER}, {OWNERCOUNT}, .IntList), OWNERCOUNT)
     => M2:IMap
-+requires:
+requires:
```
daejunpark: any specific reason for this?
it will not inherit the root's requires: `andBool #range(0 <= CD < 1024)`

denisadiaconescu: No, sorry. I fixed this small mistake.

PR #111, opened 2018-12-25
encodeTransactionData internal interface

denis-bogdanas:
Uses custom KEVM with a couple more [smtlib].

PR #122, opened 2018-12-31
Enc trans data 2

yzhang90:


daejunpark:
@yzhang90 so, you want to keep the K changes except the implication matching attribute, and also all the KEVM changes, right?

daejunpark:
subsume #111

PR #130, opened 2019-01-07
update abstraction for call

yzhang90:


gnosis/.build/.kevm.rev
```diff
@@ -1 +1 @@
-85a5342a51a238184d7876345a27e0d3b73018c8
+3f9a0176fbd1b03542618ae917197330bc0c6db2
```
yzhang90: https://github.com/kframework/evm-semantics/compare/85a5342a...3f9a0176

PR #157, opened 2019-01-20
move evm-symbolic.k from evm-semantics to VSC, make VSC work with KEVM-imap

yzhang90:
# Summary of changes
- evm-symbolic.k moved from EVM-semantics to VSC (Yi).
- Changes in all projects that use IMap to work with KEVM-imap, that is about to be merged t oKEVM-master (Yi & Denis).
- Extracted ECREC-related logic to evm-symbolic.k and ecrec-symbolic.k.
- Linked all projects that use IMap to KEVM-imap, the clean implementation of IMap about to be merged to KEVM-master.
- Linked all projects that use legacy map to the latest version of KEVM-master that is compatible with them.
- Disabled a few Gnosis tests that formerly used split() and don't work yet (but should), possibly due to issues in K.
- Linked all projects to K-master.

## State of tests and dependencies
All tests are enabled except a few that used split() before. There are 3 .build dirs:
1. All projects using imap, use KEVM-imap. (root .build)
2. Uniswap uses KEVM-uniswap.
3. All others use an older KEVM-master. (erc20/.build)

All .build/.k.rev use latest K-master.


.build/.k.rev
```diff
@@ -1 +1 @@
-7e8f009af84746f838626a206391cbbe082d52d8
+eb657b04f4979412ee0c74ac5794f94f7a974084
```
yzhang90: https://github.com/kframework/k/compare/7e8f009a...eb657b04

.build/.kevm.rev
```diff
@@ -1 +1 @@
-f0a8d473676232c404ef3439260984505922a8e5
+aca727b80e864d08cb8e799087b900ca93195fe0
```
yzhang90: https://github.com/kframework/evm-semantics/f0a8d473...aca727b8

erc20/vyper/Makefile
```diff
@@ -1,13 +1,12 @@
...
-LOCAL_LEMMAS:=../abstract-semantics.k ../verification.k
+LOCAL_LEMMAS:=../abstract-semantics.k ../verification.k ../../resources/evm-symbolic.k
 TMPLS:=../module-tmpl.k ../spec-tmpl.k
 
 SPEC_GROUP:=vyper-erc20
 SPEC_INI:=vyper-erc20-spec.ini
 
-KPROVE_OPTS:=--smt-prelude $(abspath $(dir $(abspath $(lastword $(MAKEFILE_LIST))))/../../resources/evm.smt2) \
-             --deterministic-functions --cache-func-optimized
+KPROVE_OPTS:=--smt-prelude $(abspath $(dir $(abspath $(lastword $(MAKEFILE_LIST))))/../../resources/evm.smt2)
```
denis-bogdanas: I believe we can wait until I fix K-gnosis-rebased, to avoid messing up the scripts. Whoever depends on this PR can rebase on it.

.build/.kevm.rev
denis-bogdanas: The purpose of `evm-symbolic.k` was to keep rules that 1) Are only needed for symbolic execution and 2) have to be regular rules, processed at kompile phase, that can generate branching.

All other rules can be moved to VSC, to a file that should have different name.

The only rule which I suspect HAVE to stay here is for split(). Can you explain what happened to it? I see `split` is still used in some places. You probably disabled those tests.

Until K guys decide where to incorporate it, I propose keeping it in a custom branch on KEVM, like we did so far. In this case the file in VSC should have different name, to avoid confusion.

erc20/vyper/Makefile
yzhang90: sure

.build/.kevm.rev
yzhang90: Daejun will talk to you about how to proper implement the *split* rules. The current split rule will appear neither on k-master nor on evm-master. Dwight and Everett don't like it.

PR #178, opened 2019-01-31
Jenkinsfile: limit to running on nodes labeled as proofs

ehildenb:
This should keep a single instance of this build running at a time on nodes with lots of memory.


denis-bogdanas:
jenkins: test this please

resources/kprove.mak
```diff
@@ -78,6 +78,7 @@ $(K_REPO_DIR):
 	git clone $(K_REPO_URL) $(K_REPO_DIR)
 	cd $(K_REPO_DIR) \
 		&& git reset --hard $(K_VERSION) \
+		&& git submodule update --init --recursive \
```
daejunpark: @denis-bogdanas why is this needed? we will not need llvm nor haskell backend, right?

denis-bogdanas: ^^^ @ehildenb

PR #180, opened 2019-02-05
Update k and kevm version

yzhang90:


ehildenb:
I guess this isn't needed since #178 has been merged.

PR #181, opened 2019-02-06
add -Dhaskell.backend.skip

daejunpark:


resources/kprove.mak
```diff
@@ -79,7 +79,7 @@ $(K_REPO_DIR):
 	cd $(K_REPO_DIR) \
 		&& git reset --hard $(K_VERSION) \
 		&& git submodule update --init --recursive \
```
daejunpark: is the above `git submodule update` still needed with this change?

denis-bogdanas: That line is just an experiment from Everett. I'll delete it.

denis-bogdanas:
I couldn't find the root reason why VSC on Jenkins intermittently fails. A function rule is applied that should not apply, but I cannot debug deeper than that. On my local machine it works, on Jenkins with debug options enabled it works. I don't see how this can be related to changes in K-gnosis.

Recommendation is to trigger rebuild when this happens.

PR #196, opened 2019-03-12
Makefile refactoring

denis-bogdanas:
Root Makefile refactored to make it easier to enable/disable individual projects.


Makefile
```diff
@@ -1,35 +1,63 @@
...
 
-ifneq (,$(or $(findstring all,$(MODE)),$(findstring k-test,$(MODE))))
-SUBDIRS+=k-test
+ifneq (,$(findstring dom,$(MODE)))
+SUBDIRS+=$(DOM_DIRS)
 endif
 
-ifneq (,$(or $(findstring all,$(MODE)),$(findstring erc20,$(MODE))))
-ERC20+=hkg hobby ds-token
+ifneq (,$(findstring dom,$(MODE)))
```
daejunpark: this shouldn't be `dom`

Makefile
```diff
@@ -1,35 +1,63 @@
...
+SUBDIRS+=$(CASPER_DIRS)
 endif
 
-ifneq (,$(or $(findstring all,$(MODE)),$(findstring bihu,$(MODE))))
-#SUBDIRS+=bihu
+ifneq (,$(findstring dom,$(MODE)))
```
daejunpark: this shouldn't be `dom` as well

daejunpark:
@denis-bogdanas I have a further suggestion about this refactoring.

Initially, I wrote the Makefile to allow one to specify multiple modes, like `make MODE=minimal,erc20`, but it turned out that in most cases, we use only a single mode. Also, even when we want to specify multiple modes, we can also do like, `make MODE=minimal && make MODE=erc20`. (Indeed, the only difference can be seen when we run it parallel, but the parallel run does not work well anyway.)

So, I think we can just drop the feature of specifying multiple modes, and make the Makefile much simpler.

That is, we can have only the following line, instead of the whole (verbose) if statements:
```
SUBDIRS:=$($(shell echo $(MODE) | tr a-z A-Z)_DIRS)
```

What do you think?

(I can do this as a separate PR, but I don't want to interfere your workflow for upstreaming.)

denis-bogdanas:
@daejunpark Looks nice, I'll try it. Never thought it is possible to do such perversions.

PR #199, opened 2019-03-14
tutorial+kprove.mak update

denis-bogdanas:


resources/kprove.mak
```diff
@@ -29,6 +29,13 @@ TMPLS?=module-tmpl.k spec-tmpl.k
 KPROVE_OPTS?=
 KPROVE_OPTS+=$(EXT_KPROVE_OPTS)
 
+# Define variable DEBUG to enable debug options below
+# DEBUG=true
+ifdef DEBUG
+KPROVE_OPTS+=--debug-z3-queries --log-rules \
+             --log-cells k,output,statusCode,localMem,pc,gas,wordStack,callData,accounts,#pc,#result,memoryUsed
```
daejunpark: is this duplicate with the one below?
https://github.com/runtimeverification/verified-smart-contracts/pull/199/files#diff-ec2a80a54ac1825a9fa72189f33def8fR64

daejunpark: also, you have no leading backslash before `#pc` and `#result`, while you have in the below.

denis-bogdanas: right, thanks! Wonder why it worked locally.

PR #207, opened 2019-03-24
Gnosis: fixed rules that formerly needed `split()`, to work without it

denis-bogdanas:
Now `split()` is not used by any project.

### DIFF
--- a/gnosis/Makefile
+++ b/gnosis/Makefile
@@ -31,6 +31,9 @@ SPEC_NAMES_GNOSIS_SAFE:=encodeTransactionData-public-1 \
                         checkSignatures-success \
                         checkSignatures-failure-1 \
                         checkSignatures-failure-2 \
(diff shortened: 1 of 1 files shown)


daejunpark:
what did you change? the only makefile? i don't see why the makefile change fixes the split issue. what am i missing here?

denis-bogdanas:
Yes only the makefile. Former `checkSignatures-loop-success` is now divided into checkSignatures-loop-success-middle and checkSignatures-loop-success-end, to replace the effect of `split()`. But checkSignatures-loop-success-middle needs the full checkSignatures-loop-success to pass. So I added checkSignatures-loop-success-trusted to the list of specs it needs. That's all. Most changes were done in the previous PR. I introduced checkSignatures-loop-success-middle and checkSignatures-loop-success-end there, deleted split(), but commented these new specs because they were failing or not yet tested.

PR #208, opened 2019-03-27
mandate that .k.rev is an ancestor of current master

dwightguth:


denis-bogdanas:
Does this allow referencing K non-master in: (1) branches other than VSC-master, (2) pull requests to branches other than master?

dwightguth:
I see no reason why we would need pull requests to branches other than master, and it does not affect any branch that is not being tested as part of a pull request.

PR #226, opened 2019-04-25
License under The Reproducibility License 1.1.0

kemitchell:
cc @pmackay1982 and @daejunpark 

This pull request adds a public license to the repository, which it did not have before. It also adds a new section to README mentioning that licensing, linking directly to the most important section of its text, and noting that other parts of the tool chain are and remain open source, under their respective licenses.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,65 @@
+# Reproducibility License 1.1.0
+
+## Acceptance
+
+In order to get any license under these terms, you must
+agree to their rules.  Those rules are both obligations
+under your agreement and conditions to all your licenses.
+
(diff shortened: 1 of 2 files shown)


daejunpark:
@kemitchell Does this term allows others to produce a new result based on ours for the "non-commercial" purpose?

kemitchell:
@pmackay1982 and @daejunpark, I've updated this PR with the latest license text.

dazzaji:
Any idea where this license originate from?  It's very interesting and I might want to use it for some projects that are intended to be reproducible.  Thanks.

kemitchell:
@dazzaji drop me a line at kyle@kemitchell.com.

PR #263, opened 2019-10-27
kprove.mak improvements: generalzied SPEC_GROUP, timestamp files, timeout

denis-bogdanas:


resources/kprove.mak
```diff
@@ -68,61 +84,80 @@ SPEC_FILES:=$(patsubst %,$(SPECS_DIR)/$(SPEC_GROUP)/%-spec.k,$(SPEC_NAMES))
...
-	git submodule update --init -- $(PANDOC_TANGLE_SUBMODULE)
+$(KEVM_REPO_DIR):
+	git clone $(KEVM_REPO_URL) $(KEVM_REPO_DIR)
 
 #
 # Specs
 #
 
-split-proof-tests: $(SPECS_DIR)/$(SPEC_GROUP) $(SPECS_DIR)/lemmas.k $(SPEC_FILES)
+split-proof-tests: $(SPECS_DIR)/$(SPEC_GROUP)/lemmas.timestamp $(dir $(SPECS_DIR)/$(SPEC_GROUP))/lemmas.k $(SPEC_FILES)
```
daejunpark: `$(SPECS_DIR)/lemmas.k` is also needed, no?

resources/kprove.mak
```diff
@@ -36,28 +54,26 @@ ifdef DEBUG
...
 K_REPO_DIR:=$(abspath $(BUILD_DIR)/k)
 KEVM_REPO_DIR:=$(abspath $(BUILD_DIR)/evm-semantics)
 
 K_BIN:=$(abspath $(K_REPO_DIR)/k-distribution/target/release/k/bin)
 
-KPROVE:=$(K_BIN)/kprove -v --debug -d $(KEVM_REPO_DIR)/.build/defn/java -m VERIFICATION --z3-impl-timeout 500 \
+TIMEOUT_PREFIX:=timeout -s SIGINT -k 10s --foreground $(TIMEOUT)
+
+KPROVE:=$(TIMEOUT_PREFIX) $(K_BIN)/kprove -v --debug -d $(KEVM_REPO_DIR)/.build/defn/java -m VERIFICATION \
+        --z3-impl-timeout 500 --shutdown-wait-time 5000 \
```
daejunpark: Is `--shutdown-wait-time 5000` a new option added to K?

resources/kprove.mak
denis-bogdanas: No, the line above includes this most common case. Otherwise jenkins would have failed.

resources/kprove.mak
denis-bogdanas: No, it's old. But it only makes sense in the context of SIGINT (or ctrl+C). That's why I added it now. When kprove gets SIGINT, it interrupts the main thread and waits for 5 seconds until main threads propagates the exception and prints all it wants. Whether this really happens depends on what kprove was doing at the time of interruption.

denis-bogdanas:
I'll add option to disable timeout in another PR.

PR #275, opened 2019-11-26
Deposit

daejunpark:


deposit/bytecode-verification/lemmas.k
```diff
@@ -0,0 +1,672 @@
+requires "edsl.k"
```
denis-bogdanas: Keeping multiple versions of lemmas.k/lemmas.md is going to be hard to maintain. Can you use the common lemmas.md and import all the rest from verification.k? Or else, maybe we should discuss what's the difference between new lemmas structure and old one. Maybe you want to use lemmas.md for old projects only. I have no way to know.

deposit/bytecode-verification/lemmas.sh
```diff
@@ -0,0 +1,21 @@
...
+    imports EDSL
+EOF
+
+cat lemmas/macro.k \
+    lemmas/range.k \
+    lemmas/arith.k \
+    lemmas/buf.k \
+    lemmas/map.k \
+    lemmas/hash.k \
+    lemmas/gas.k
```
denis-bogdanas: Why not just importing them all from verification.k or a similar module and avoid extra scripts?

deposit/bytecode-verification/run.sh
```diff
@@ -0,0 +1,105 @@
...
+      --concrete-rules "EVM-DATA.take.zero-pad" \
+      `eval echo $1` \
+      >$LOGDIR/`basename $1`.log 2>&1
+    tail -100 $LOGDIR/`basename $1`.log
+  }
+  export -f run_kprove
+
+  echo $SPECS | xargs -d ' ' -n 1 -P 2 -I {} bash -c 'run_kprove "{}"'
+
+  date
```
denis-bogdanas: I guess the diff between this script and makefile is:
1. Extra command options.
2. Different set of specs to run.
3. the `tail` command.

I'd recommend moving all what is possible to Makefile, to keep minimalistic script structure. First 2 features can be implemented in makefile as extra options. The 3rd also can be done in Makefile as alternative command compared to `make test` or whatever is default. I can help if you want.

deposit/bytecode-verification/lemmas.k
daejunpark: As discussed offline, this module is written for a better organization of lemmas, facilitating an external review, at the cost of duplication (of lemmas from `lemmas.md` and `evm-data-map-symbolic.k`). See the comment in the file.

deposit/bytecode-verification/Makefile
```diff
@@ -0,0 +1,68 @@
+BUILD_DIR:=.build
+
+LOCAL_LEMMAS:=verification.k \
+			  abstract-semantics.k \
```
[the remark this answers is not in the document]
daejunpark: fixed

deposit/bytecode-verification/lemmas.sh
daejunpark: dropped

PR #280, opened 2019-12-03
deposit: fix semantics rule duplication

daejunpark:


deposit/bytecode-verification/Makefile
```diff
@@ -1,3 +1,6 @@
+K_VERSION_FILE   :=.build/.k.rev
```
denis-bogdanas: The point of this separation was to have one version of K for everything, unless some project really needs a custom version. This reduces build times and will make it easier use kserver in future if needed. You can just update global .k.rev Existing version in .k.rev already has `--concrete-rules` btw. Please note that my commit merged to K-master (the one pointed by this commit) is just a refactoring, no performance improvements yet. I'll make a PR when the piece that adds the definition cache gets merged to K-master.

So either change global .k.rev or remove this change.

daejunpark: removed

PR #288, opened 2019-12-09
Deposit lemmas

MrChico:
Reduces the trusted base of the deposit contract verification, mainly by removing lemmas that I think are not used by these specs anyway.

### DIFF
--- a/deposit/bytecode-verification/lemmas.k
+++ b/deposit/bytecode-verification/lemmas.k
@@ -83,12 +83,6 @@ module LEMMAS
     rule 0 <=Int #asWord(WS)             => true
     rule         #asWord(WS) <Int pow256 => true
 
-    rule 0 <=Int hash1(_)             => true
-    rule         hash1(_) <Int pow256 => true
-
-    rule 0 <=Int hash2(_,_)             => true
-    rule         hash2(_,_) <Int pow256 => true
-
     rule 0 <=Int chop(V)             => true
     rule         chop(V) <Int pow256 => true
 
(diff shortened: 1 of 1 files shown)


deposit/bytecode-verification/lemmas.k
```diff
@@ -182,21 +176,21 @@ module LEMMAS
     rule N &Int MASK => N  requires MASK ==Int (2 ^Int (log2Int(MASK) +Int 1)) -Int 1 // MASK = 0xffff...f
                             andBool 0 <=Int N andBool N <=Int MASK
 
-    // 0xff..ff00..00 (16 1's followed by 240 0's)
-    rule 115790322390251417039241401711187164934754157181743688420499462401711837020160 &Int #asWord(WS1 ++ WS2)
+    // 0xff00..00 (16 1's followed by 240 0's)
+    rule (pow256 -Int (2 ^Int 240)) &Int #asWord(WS1 ++ WS2)
```
daejunpark: This does not work as expected. Function symbols are not evaluated in the LHS of function rules, only do macros. The fact that proofs pass, this lemma is not used at all.

deposit/bytecode-verification/lemmas.k
```diff
@@ -182,21 +176,21 @@ module LEMMAS
...
         => #asWord(WS1 ++ #buf(30, 0))
       requires #sizeWordStack(WS1) ==Int 2
        andBool #sizeWordStack(WS2) ==Int 30
 
     // x &Int (NOT 31)
-    rule X &Int 115792089237316195423570985008687907853269984665640564039457584007913129639904 => (X /Int 32) *Int 32  requires 0 <=Int X
+    rule X &Int (pow256 -Int 32) => (X /Int 32) *Int 32  requires 0 <=Int X
```
daejunpark: Same above.

deposit/bytecode-verification/lemmas.k
```diff
@@ -182,21 +176,21 @@ module LEMMAS
...
 
     // 2^256 - 2^160 = 0xff..ff00..00 (96 1's followed by 160 0's)
-    rule 115792089237316195423570985007226406215939081747436879206741300988257197096960 &Int ADDR => 0
+    rule (pow256 -Int pow160) &Int ADDR => 0
```
daejunpark: Same above.

daejunpark:
see this: https://github.com/runtimeverification/verified-smart-contracts/commits/deposit

daejunpark:
Subsumed by #289

PR #289, opened 2019-12-10
Cleanup Deposit Lemmas

daejunpark:


daejunpark:
Subsume #288

daejunpark:
> Can any of these simplifications be migrated to ERC20?

Yes, see #290.

PR #316, opened 2020-02-10
more comments on memory addresses

daejunpark:


deposit/bytecode-verification/deposit-spec.ini
```diff
@@ -55,55 +55,55 @@ VYPER_GENERATED_BOUNDS:
...
-    [ 320 := #buf(32, DEPOSIT_COUNT) ]  /* argument */
-    [ 384 := #buf(32, Y8) ]             /* y: uint256 = (y << 8) + bitwise_and(x, 255) */
-    [ 416 := #buf(32, X8) ]             /* x: uint256 = (x >> 8) */
-    [ 448 := #buf(32, 8) ]              /* for-loop index */
-    [ 480 := #buf(32, X7 &Int 255) ]    /* bitwise_and(x, 255) */
-    [ 544 := #buf(32, Y8) ]             /* memcpy(544 <- 384, 32) */
-    [ 536 := #buf(32, 8) ]              /* for slice(convert(y, bytes32), start=24, len=8) */
-    [ 704 := #buf(32, 8) ++ #bufSeg(#buf(32, Y8), 24, 8) ]  /* return: bytes[8] */ /* memcpy(704 <- 536, 40) */
+    [ 352 /* = 320 + 32 *  1      */ := #buf(32, {RETURN_ADDR_FROM_TO_LITTLE_ENDIAN_64}) ]  /* return address */
+    [ 320 /* = 320 + 32 *  0      */ := #buf(32, DEPOSIT_COUNT) ]  /* argument */
```
MrChico: any reason why these are out of order?

deposit/bytecode-verification/deposit-spec.ini
```diff
@@ -55,55 +55,55 @@ VYPER_GENERATED_BOUNDS:
...
-    [ 762 <- 0 ]
-    [ 763 <- 0 ]
-    [ 764 <- 0 ]
-    [ 765 <- 0 ]
-    [ 766 <- 0 ]
-    [ 767 <- 0 ]
-    [ 672 := #buf(32, 32) ]             /* ??? */
-    [ 640 := #buf(32, 0) ]              /* ??? */
+    [ 800 /* = 320 + 32 * 15      */ := #buf(32, 32) ]             /* padding loop index */
+    [ 744 /* = 320 + 32 * 13 +  8 */ <- 0 ]                        /* padding */
```
MrChico: `[ 744 /* = 320 + 32 * 13 +  8 */ := #buf(22, 0) ]` would rewrite to this, no? That seems cleaner to me

MrChico:
I noticed you assume `memUsed` to be 0 in the rhs of `to_little_endian_64` spec. This should probably be abstracted to some `MU` instead, since this is an internal function

deposit/bytecode-verification/deposit-spec.ini
daejunpark: The main reason is that K does not yet fully support the AC matching over maps.  So, I put the map update in the order they appear in the execution of the bytecode.  The advantage of this order is that it clearly shows how the local memory is updated in each step of the bytecode execution.

deposit/bytecode-verification/deposit-spec.ini
daejunpark: Yes, they are equivalent, but the current form is easier for K to prove.

daejunpark:
@MrChico You're right. But note that since no `MSIZE` opcode appears in the bytecode, the only effect of having it to be 0 in the LHS is that the `GAS_COST` value of the specification is an over-approximation of the actual cost. The precise (symbolic) expression of the gas cost for the entire contract was not in the scope.

> I noticed you assume `memUsed` to be 0 in the rhs of `to_little_endian_64` spec [here](https://github.com/runtimeverification/verified-smart-contracts/blob/master/deposit/bytecode-verification/deposit-spec.ini#L315). This should probably be abstracted to some `MU` instead, since this is an internal function

daejunpark:
@MrChico I'll merge this PR for now, as your questions are not necessarily relevant to the changes in this PR.  Please open github issues if your questions are not fully addressed by my answers, so that we can discuss further.  Thanks for your detailed review!

PR #317, opened 2020-03-18
Deposit: Improve init spec

daejunpark:
Now the `init` spec becomes straightforward, specifying only two cases: success and failure. The success case spec describes the entire function execution, specifying only the status code, the output, the storage update, and the gas cost.

More spec improvement will come, and the original full spec is moved to `deposit-spec-full.ini` for a backup and a future reference.

### DIFF
--- a/deposit/bytecode-verification/Makefile
+++ b/deposit/bytecode-verification/Makefile
@@ -12,10 +12,8 @@ KPROVE_OPTS:=--smt-prelude $(abspath $(dir $(abspath $(lastword $(MAKEFILE_LIST)
              --branching-allowed 0 \
              --z3-cnstr-timeout 100
 
-SPEC_NAMES:=init-init \
-            init-loop0 \
-            init-loop-enter \
-            init-loop-exit \
+SPEC_NAMES:=init-success \
+            init-revert \
             to_little_endian_64 \
             get_deposit_root-init \
             get_deposit_root-loop0-then \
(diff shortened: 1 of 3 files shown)


daejunpark:
Also the get_deposit_count specs are simplified, where the local memory update details are omitted and the trusted spec of to_little_endian_64 is no longer used.

The performance comparison w/ vs w/o the trusted spec of to_little_endian_64:

With to_little_endian_64 trusted spec:
```
Execution paths: 1
Longest path: 10157 steps
Stats for each phase, time, used memory, implicit main GC time percentage:
Total                 :  239.138 s,	 1193 MB, gc:  1.366 %
  JVM init            :    0.038 s
  Parsing             :  149.149 s,	 1454 MB, gc:  1.213 %
  Rewriter init       :    2.749 s,	 1035 MB, gc: 20.517 %
  Execution           :   87.202 s,	 1193 MB, gc:  1.025 %
```

Without to_little_endian_64 trusted spec:
```
Execution paths: 1
Longest path: 28200 steps
Stats for each phase, time, used memory, implicit main GC time percentage:
Total                 :  343.026 s,	 2441 MB, gc:  1.787 %
  JVM init            :    0.057 s
  Parsing             :  126.541 s,	 1219 MB, gc:  1.164 %
  Rewriter init       :    1.299 s,	  858 MB, gc:  0.000 %
  Execution           :  215.128 s,	 2441 MB, gc:  2.165 %
```

The parsing time is slightly reduced (149s -> 126s), as the trusted spec is not included. The execution time is increased to 2.5x longer (87s -> 215s), as the number of execution steps is increased to 2.7x bigger (10,157 -> 28,200 steps). Although the verification time is increased, the spec becomes simpler and it seems worth paying for the succinctness.

MrChico:
Nice! This looks good to me!

PR #318, opened 2020-03-19
Deposit: simplify get_deposit_root spec

daejunpark:
This simplifies the get_deposit_root spec. Only the necessary part (i.e., the local variables) of the local memory update is specified. Unlike the init spec, it is not feasible to explicitly enumerate all paths of the function, because although the loop bound is fixed to 32, it leads to 2^32 paths, due to the if-statement in the loop body. Thus, we adopt the refinement-based approach (described [here](https://github.com/runtimeverification/verified-smart-contracts/wiki/Our-Refinement-Based-Formal-Verification-Approach)) to write a spec recursively.

### DIFF
--- a/deposit/bytecode-verification/deposit-spec.ini
+++ b/deposit/bytecode-verification/deposit-spec.ini
@@ -53,6 +53,12 @@ VYPER_GENERATED_BOUNDS:
     [  96 := #buf(32, 115792089237316195423570985008687907853099843482180094807725896704197245534208) ]
     [ 128 := #buf(32, 1701411834604692317316873037158841057270000000000) ]
     [ 160 := #buf(32, 115792089237316195423570985006986496018665292348323691002298742950633129639936) ]
+ENSURES_VYPER_GENERATED_BOUNDS:
+    andBool #range(NEW_MEM,  32, 32) ==K #buf(32, 1461501637330902918203684832716283019655932542976)
+    andBool #range(NEW_MEM,  64, 32) ==K #buf(32, 170141183460469231731687303715884105727)
+    andBool #range(NEW_MEM,  96, 32) ==K #buf(32, 115792089237316195423570985008687907853099843482180094807725896704197245534208)
+    andBool #range(NEW_MEM, 128, 32) ==K #buf(32, 1701411834604692317316873037158841057270000000000)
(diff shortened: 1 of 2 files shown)


deposit/bytecode-verification/deposit-spec.ini
```diff
@@ -407,181 +415,98 @@ storage: M
...
-    [ 448 /* = 320 + 32 *  4 */ := #buf(32, ANON_5) ]     /* size of sha256 input */
-    [ 480 /* = 320 + 32 *  5 */ := #buf(32, ANON_6) ]     /* node */
-    [ 512 /* = 320 + 32 *  6 */ := #buf(32, ANON_7) ]     /* zero_hashes[height] */
+    [ 320 /* = 320 + 32 * 0 */ := #buf(32, 0) ]          /* zero_bytes32: bytes32 = 0x0 */
+    [ 352 /* = 320 + 32 * 1 */ := #buf(32, NODE) ]       /* node: bytes32 */
+    [ 384 /* = 320 + 32 * 2 */ := #buf(32, SIZE) ]       /* size: uint256 */
+    [ 416 /* = 320 + 32 * 3 */ := #buf(32, HEIGHT) ]     /* height */
 +requires:
     // conditions
     andBool #range(0 <= HEIGHT <= 32)
```
MrChico: Is there a spec that ensures that this reverts if this is not fullfilled? Perhaps `get_deposit_root-revert` can be extended to deal with this case as well

daejunpark: Indeed this is a loop invariant that can be proved (using the refinement approach mentioned above), thus it is not possible to violate this. If this is violated somehow (e.g., due to a VM bug), it does not revert but ends up an infinite loop (because the loop termination condition is `height =/= 32` not `height > 32`).

daejunpark: I clarified this in the spec: https://github.com/runtimeverification/verified-smart-contracts/pull/320/commits/64b6810fab85582628887fb78b18a745e1e3a131

daejunpark:
The diff is confusing, and it would be easier to just review the new spec:
https://github.com/runtimeverification/verified-smart-contracts/blob/deposit/deposit/bytecode-verification/deposit-spec.ini#L399-L527

PR #319, opened 2020-03-20
Deposit: simplify deposit spec

daejunpark:
This simplifies the deposit spec, in a similar way to the previous simplification (#317 and #318).

- Now we have a single spec `[deposit-success-data]` that specifies the path from the beginning of the function to the beginning of the loop (i.e., [lines 71-99](https://github.com/ethereum/eth2.0-specs/blob/v0.11.0/deposit_contract/contracts/validator_registration.vy#L71-L99)). That spec does not include the trusted to_little_endian_64 spec, as done in #317. 

- The loop specs are also simplified, similarly to #318.

### DIFF
--- a/deposit/bytecode-verification/Makefile
+++ b/deposit/bytecode-verification/Makefile
@@ -12,47 +12,44 @@ KPROVE_OPTS:=--smt-prelude $(abspath $(dir $(abspath $(lastword $(MAKEFILE_LIST)
              --branching-allowed 0 \
              --z3-cnstr-timeout 100
 
-SPEC_NAMES:=init-success \
-            init-revert \
-            to_little_endian_64 \
-            get_deposit_root-success-loop1-then \
-            get_deposit_root-success-loop1-else \
-            get_deposit_root-success-loop-body-then \
-            get_deposit_root-success-loop-body-else \
-            get_deposit_root-success-loop-exit \
-            get_deposit_root-revert \
-            get_deposit_count-success \
-            get_deposit_count-revert \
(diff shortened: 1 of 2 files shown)


daejunpark:
Performance comparison between with vs without the trusted spec.

With the trusted spec:
```
Execution paths: 1
Longest path: 58610 steps
Stats for each phase, time, used memory, implicit main GC time percentage:
Total                 : 1013.714 s,	 4296 MB, gc:  1.631 %
  JVM init            :    0.034 s
  Parsing             :  133.969 s,	 1765 MB, gc:  1.993 %
  Rewriter init       :    1.921 s,	 1429 MB, gc:  0.000 %
  Execution           :  877.790 s,	 4296 MB, gc:  1.579 %
```

Without the trusted spec:
```
Execution paths: 1
Longest path: 94696 steps
Stats for each phase, time, used memory, implicit main GC time percentage:
Total                 : 1391.843 s,	 5830 MB, gc:  1.579 %
  JVM init            :    0.038 s
  Parsing             :  104.690 s,	 1921 MB, gc:  1.459 %
  Rewriter init       :    1.492 s,	 1286 MB, gc:  0.000 %
  Execution           : 1285.623 s,	 5830 MB, gc:  1.591 %
```

daejunpark:
The diff doesn't help much. It would be easier to review the new spec: https://github.com/runtimeverification/verified-smart-contracts/blob/deposit2/deposit/bytecode-verification/deposit-spec.ini#L403-L764

PR #320, opened 2020-03-20
Deposit: further improvement of specs

daejunpark:
This further improves the specs.
- Now the init spec considers arbitrary accidental and/or malicious calldata garbage for security.
- The loop invariant regarding the loop index for both get_deposit_root and deposit is clarified.
- The gas/memory cost description is simplified to avoid duplication in the ini file.
- The unnecessary cells are omitted in the specification.
- And, some minor reorgs.


daejunpark:
Ignore the first three commits, which will be rebased once #318 and #319 are merged.

